PR #5, opened 2018-11-14
Impl Debug for InsertError

pablostebler:


vberger:
Could you just add a line in the changelog for these addition, please?

vberger:
Hmm, thinking back about this, while I'm absolutely in favor of implementing `Debug` for the error types, I'm not sure this is relevant for `Generic<S>`.

I'd probably rather have a manual implementation of `Debug` for `InsertError<E>` that does not depend on `E` being `Debug`.

vberger:
Perfect, thanks!

PR #13, opened 2020-04-19
Rework calloop without mio

### LINKED ISSUE #7: Ownership issues with LoopHandle::insert_source
`LoopHandle::insert_source` takes ownership of `E` and moves it to the return value.  This behavior is undocumented and might be confusing.  A minimal example:

```rust
use calloop::EventLoop;
use calloop::signals::{Signal, Signals};
use std::process;

fn main() {
    println!("pid: {}", process::id());
    let mut event_loop = EventLoop::<()>::new().unwrap();
    let signals = Signals::new(&[Signal::SIGUSR1]).unwrap();
    handle.insert_source(signals, |_, _| {
        println!("SIGUSR1");
    }).unwrap();
    loop {
        event_loop.dispatch(None, &mut ());
    }
}
```

Expected behavior: Prints "SIGUSR1" upon receiving SIGUSR1.
Actual behavior: The process terminates at SIGUSR1.  Since `signals` has been moved to the return value of `insert_source` and dropped, the signals are unmasked.

The fix here is to bind the return value of `insert_source`.  I think this is rather counter-intuitive.  Perhaps `insert_source` should borrow the source instead?

### LINKED ISSUE #12: Rethinking calloop
Over the last few days it appeared that `mio`'s behavior, and notably its focus on edge-triggered events, is an undesirable property. And as such, it may be relevant to drop it from being a calloop dependency. That would also mean focusing calloop on unix platforms, which I'm pretty confident it is only used on anyway.

On this process, I'm considering re-shaping calloop to, at the same time, fix the long-stating issue of ownership weirdness regarding the event sources. This issue is a design draft to gather feedback.

-------

Internally, calloop would contain a small abstraction around epoll and the appropriate equivalent for other platforms. This is a mostly-internal API, and only relevant for users that are going to implement their own event sources.

```rust
pub struct Poll { /* ... */ }

impl Poll {
    /// Register a new FD to the poll structure
    pub fn register(&mut self,
        fd: RawFd, /* the FD to register for polling */
        interest: Interest, /* Read / Write / Other... bitflag */
        mode: PollMode, /* oneshot, level, edge */
        token, Token /* opaque token to recognize the source */
    -> io::Result<()> {
        /* ... */
    }

    /// Update an FD registration
    pub fn update_registration(&mut self,
        fd: RawFd, i: Interest, m: PollMode, t: Token
    ) -> io::Result<()> {
        /* ... */
    }

    /// Remove a redigstration
    pub fn unregister(&mut self, fd: RawFD) -> io::Result<()> {
        /* ... */
    }
}
```

Now, comes the central trait of the crate: `Source`, representing an event source:

```rust
pub trait Source {
    /// The type of events this event source generates
    type Event;
    /// The return type expected from the callback closure, if the user
    /// needs to communicate back to the source.
    /// Set to () is not needed.
    type Ret;

    /// Process readiness of the source and translate it as events to the
    /// provided closure
    fn process_events<F>(&mut self, readiness: Readiness, callback: F) -> io::Result<()>
        where F: FnMut(Self::Event) -> Self::Ret;

    /// Make the source register itself to the `Poll`
    pub fn register(&mut self, &mut Poll, token: Token) -> io::Result<()> {
        /* ... */
    }

    /// Make the source update its registration
    pub fn update_registration(&mut self, &mut Poll, token: Token) -> io::Result<()> {
        /* ... */
    }

    /// Unregister the source
    pub fn unregister(&mut self, &mut Poll) -> io::Result<()> {
        /* ... */
    }
}
```

From this, the calloop user, to insert a new event source in the loop, needs to provide an implementation of the `Source` trait and a closure with matching signature. The loop takes ownership of both and provide a `SourceToken<S>` in exchange.

This `SourceToken<S>` can be usedto remove  the source from the `EventLoop`. Requires the same access as to insert a new source (so likely a `LoopToken`), the user is given the source `S` back. The closure is dropped.

Dropping the `SourceToken<S>` does nothing.

-------

Remaining questions:

- Is it worth introducing a `Waker` mechanism to remotely wake-up the event loop?

I'm thinking a special `Source` that just generates `()` events and can be remotely made ready should do that well, without requiring a specific `Waker` mechanism.

- This API draft allows a source to register more than one FD to the poll instance, however it does not allow the source to know which FD was made ready. Should we handle that?

A possibility would be to introduce a notion of sub-token (by splitting the u64 token into two u32s for example). An other would be to provide a `Multiplexer` source, which can aggregate several event sources by internally creating its own epoll instance and nesting them.

- Is it worth giving remote control of the source via the `SourceToken`? By allowing it to distantly disarm/rearm an event source for example, or destroy it (so that it is dropped) ?

vberger:
The changes introduces are, for the main part:

- Drop `mio` and create of the `sys` module containing the `Poll` type, which represents the polling system used. For now only epoll is implemented, others (like kqueue) should be added for other unix platforms.
- Re-architecture of the event sources, so that the `EventLoop` actually takes ownership of the event sources. This drastically simplifies the internal logic and implementation of other event sources
- Introduce the `Ping` event source, which allows to generate `()` event. Used as a building block by `Channel` and `Timer` to replace the `mio::Waker`.
- Event sources can now register more than one FD to the event queue, and use the `sub_id` field of `Token` to differentiate between them.

Closes #12 
Closes #7


src/sys/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+
+/// A Token for registration
+///
+/// This token is given to you by the event loop, and you should
+/// forward it to the `Poll` when registering your file descriptors.
+///
+/// It also contains a publc field that you can change. In case your event
+/// source needs to register more than one FD, you can register each with a
+/// different value of the `sub_id` field, to differentiate them. You can then
+/// know which of these FD si ready by reading the `sub_id` field of the
```
Drakulix, line 68: Typo

PR #18, opened 2020-05-08
Allow non 'static Data

fengalin:
The dispatched `Data` is not captured in the `callback`, so it should not be constrained to be `'static`.

See new test [non_static_data](https://github.com/fengalin/calloop/blob/ab0cfaff22c0478dffd59e4f86c50cddb7fdc235/src/loop_logic.rs#L747).

I think we could also "loosen" the constraint on the `EventSource` & `callback` functions so that their lifetime be bound to the `EventLoop`'s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- Allow non-`'static` dispatch `Data`. `Data` is passed as an argument to the
+  `callback`s while dispatching. This change allows defining `Data` values which
+  can live no longer than the `dispatch` call and which type can hold references
+  to other values.
+
 ## 0.6.2 -- 2020-04-23
 
 - Update the README and keywords for crates.io
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -7,10 +7,10 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use crate::list::SourceList;
-use crate::sources::{EventSource, Idle, Source};
+use crate::sources::{EventSource, Idle, IdleDispatcher, Source};
 use crate::Poll;
 
-type IdleCallback<Data> = Rc<RefCell<Option<Box<dyn FnMut(&mut Data)>>>>;
+type IdleCallback<Data> = Rc<RefCell<dyn IdleDispatcher<Data>>>;
 
 struct LoopInner<Data> {
     poll: RefCell<Poll>,
@@ -58,7 +58,7 @@ impl<E> From<InsertError<E>> for io::Error {
     }
 }
 
-impl<Data: 'static> LoopHandle<Data> {
+impl<Data> LoopHandle<Data> {
     /// Insert an new event source in the loop
     ///
     /// The provided callback will be called during the dispatching cycles whenever the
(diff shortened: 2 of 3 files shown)


vberger:
Oh, nice, that's interesting. I didn't realize this requirement could be lifted with so little changes, thanks!

I don't have much to add, it looks good. :+1: Could you just add a changelog entry for it?

fengalin:
Updated the followings:

- Changed the test to exemplify the "allows defining `Data` values which can live no longer than the `dispatch` call" part.
- Updated the changelog.
- I also added a commit for a renaming proposal. Feel free to remove it if you don't think it's worth it!

A bit of context: I started updating `winit` to latest `smithay-client-toolkit`, so I had to remove the constraint on `Data` to honour the non-static `callback` constraint in `winit::platform::desktop::EventLoopExtDesktop::run_return`. I'm pretty close to having a working draft, but I may have some questions about handling some `Pointer` & `AutoThemer` stuff later.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## 0.6.3 -- 2020-05-08
```
vberger, line 5: Please don't add this line, and keep this change in the `Unreleased` section, I add the line when I actually do the release. :)

fengalin: Of course! :)

vberger:
Alright, nice (apart for the small changelog nit) :+1: I'm good with the rename, it's an internal type anyway.

BTW, regarding winit upgrade, are you aware of https://github.com/rust-windowing/winit/pull/1544 ?

fengalin:
> BTW, regarding winit upgrade, are you aware of [rust-windowing/winit#1544](https://github.com/rust-windowing/winit/pull/1544) ?

Oh no, I wasn't!

vberger:
Thanks!

> I think we could also "loosen" the constraint on the EventSource & callback functions so that their lifetime be bound to the EventLoop's.

Do you already have a precise idea of how this could be done?

fengalin:
> Do you already have a precise idea of how this could be done?

I do but I still need to try it. Will open a PR if it turns out to be applicable.

Matthias-Fauconneau:
I also had the issue passing non static DispatchData and was erasing the lifetime.
Since I can't imagine any application where 'static is not too constraining, I assumed it was an intentional early type erasure optimisation to avoid some codegen issues 😑.

Oh no! I hope you didn't spend too much time on the port. I spent a lot of time thinking I'm nearly there until stumbling on an issue requiring to redesign everything 🙄.

Did you follow in the same direction or found anything new ?
Don't hesitate to ask for any clarification :)

https://github.com/Matthias-Fauconneau/winit/blob/85ca15b1ad268f17525043b84050e0feb407cd1f/src/platform_impl/linux/wayland/event_loop.rs#L114

fengalin:
I moved to other stuffs when I realized you were already looking at this. I'd be happy to help though. I'll update my winit WIP branch with the newest calloop changes in the coming days to check if it leads somewhere and will take a look at your branch.

What issue did you stumble on which would require to redesign everything?

Matthias-Fauconneau:
I'm not working on it anymore.
It compiled but failed a display test.
I basically rewrote the backend because it was a mess of mutex arcs deadlocking on reentrancy.
I'd be glad if someone could complete testing and review :)
The new design uses DispatchData to pass the ref/mut data needed to handle the callbacks

fengalin:
Ok! Could you clean up the commit stack of your branch so I will start from it then?

Matthias-Fauconneau:
Never did this.
Can you flatten it ?

P.S: Probably better to reply on the PR ;) maybe also IM ?

PR #20, opened 2020-05-28
Allow non 'static Sources and callbacks...

fengalin:
... so they can hold references to other values.

**Breaking Changes**:

- `LoopHandle::insert_source` now returns the `Token` which can be used to
  `enable`, `disable`, `update`, ... the source just like before.
- `LoopHandle::with_source` was removed (see below).
- `LoopHandle::remove` doesn't return the event source any more (see below).
- `LoopHandle::register_dispatcher` can be used in place of
  `LoopHandle::insert_source` when the source needs to be accessed after
  its insertion in the loop.

I had to expose `Dispatcher` and proceed with the above breaking changes
because previous mechanism relied on `Any` to restore the concrete `EventSource`
and this implied a `'static` lifetime.

I tried to stick to the existing API as much as possible though. See the tests for
the changes from a user's perspective.


vberger:
Oh wow, thanks for your effort!

I agree with the general intent of your PR, but I'll need to review it in detail to ensure I'm ok with everything. Sadly I don't have much time to do that right now, so this will have to wait a few days, sorry for that.

fengalin:
No problem :)

src/loop_logic.rs
```diff
@@ -113,124 +140,98 @@ impl<Data> LoopHandle<Data> {
...
-            .downcast::<S>()
-            .map(|s| *s)
-            .unwrap_or_else(|_| panic!("Downcasting failed!"))
     }
 
-    /// Remove this event source from the event loop and drop it
+    /// Removes this source from the event loop.
     ///
-    /// The source is not given back to you but instead dropped
+    /// The source is removed but NOT unregistered.
```
vberger, line 207: The loop dispatching logic contains code to unregister killed sources (line 270), unless this is no longer valid?

src/loop_logic.rs
```diff
@@ -113,124 +140,98 @@ impl<Data> LoopHandle<Data> {
...
     ping: crate::sources::ping::Ping,
 }
 
-impl<Data> EventLoop<Data> {
+impl<'l, Data> EventLoop<'l, Data> {
     /// Create a new event loop
     ///
     /// Fails if the initialization of the polling system failed.
-    pub fn new() -> io::Result<EventLoop<Data>> {
+    // FIXME try_new?
```
vberger, line 233: :+1:

src/sources/mod.rs
```diff
@@ -144,34 +133,81 @@ pub(crate) trait EventDispatcher<Data> {
...
+
+    fn reregister(&self, poll: &mut Poll, token: Token) -> io::Result<()>;
+
+    fn unregister(&self, poll: &mut Poll) -> io::Result<()>;
+}
+
+/// An event source with its callback.
+///
+/// The `Dispatcher` can be registered in an event loop.
+/// Use the `as_source_ref_{ref,mut}` functions to interact with the event source.
```
vberger, line 146: I think there is one `_ref` too many here :wink:

vberger (changes requested):
Thank you for your PR!

I am terribly sorry about how long it took me to come back with a feedback in this. :disappointed: 

So, overall I like the design direction you've chosen, I noted a few small things (see the other review messages).

I just don't think it's appropriate to use `Token` as a type for identifying the sources for the `LoopHandle` methods. `Token` is cloneable, which could bring problems (what if the user keeps around an old token corresponding to a removed source, and tries to use it with `LoopHandle` ?). So I think I'd prefer to keep `Source` around, even just as an opaque non-`Clone` wrapper around a `Token`.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+- Allow non-`'static` event sources and callbacks, so they can hold references
+  to other values. **Breaking Changes**:
+  - `LoopHandle::insert_source` now returns the `Token` which can be used to
+    `enable`, `disable`, `update`, ... the source just like before.
+  - `LoopHandle::with_source` was removed (see below).
+  - `LoopHandle::remove` doesn't return the event source any more (see below).
+  - `LoopHandle::register_dispatcher` can be used in place of
+    `LoopHandle::insert_source` when the source needs to be accessed after
```
kchibisov, line 12: Since API is changing to use a dispatcher, instead of `with_source`, I'd encourage to write more verbose changelog on that and not `(see below)`, since it's a bit unclear about it here.

fengalin:
Thank you for reviewing. I'll try to address the changes next week.

vberger:
Hey @fengalin,

I'm planning some changes to calloop, in link with #27, to introduce a way to wake-up IO futures powered by calloop. I suspect this will touch a significant portion of calloop's internals, as such I think it'd be nicer to have your PR merged before that.

Would you have some time to come back to it soonish? If not, would you be okay with me doing the finishing touches on your PR myself?

fengalin:
Hey @vberger! Sorry for the delays, I'm getting back to it.

fengalin:
> I just don't think it's appropriate to use Token as a type for identifying the sources for the LoopHandle methods. Token is cloneable, which could bring problems (what if the user keeps around an old token corresponding to a removed source, and tries to use it with LoopHandle ?). So I think I'd prefer to keep Source around, even just as an opaque non-Clone wrapper around a Token.

I think the reason why I started using `Token` was that it seemed confusing to me to use "source" for both the registration result and for `EventSource`. Also note that the comment for the `Source` type is "A token representing an event source inserted in the event loop".

I propose using `Token` and not deriving `Clone` on it for the public registration result. This way we can consume the `Token` when user calls `remove`. We could also constrain the `Token` lifetime so that it doesn't live longer than its `LoopInner`. I would introduce a `Clone`-able `pub(carte)` `TokenInner` or `TokenPriv` for internal uses requiring cloning the `Token`. What do you think?

vberger:
The thing is that `Token` is already aprt of the public API, as manipulating it is required for implementing the `EventSource` trait for new event sources. And in this context, it needs to be `Clone`-able, for example for event sources that aggregate multiple inner event sources.

So I'm confident we need two different "token" types, one for use by implementations of the trait, and one for representing sources inserted in the loop. The first being Clone/Copy, the second not. I don't have very strong feeling about their names though, so if you think `Token` should be used for the second and you have a better name for the first, I'm all ears.

I'm not sure about the lifetime restriction, it seems to me it would make them much more complex to store around.

fengalin:
Indeed, I forgot about `Token` being used in `Source`s impl.

I guess I'll go with `RegistrationToken` for now and see if I come up with something better. I'd like to take a couple of hours to make sure the `Dispatcher` lifetimes definition is not too restrictive for some corner cases.

vberger:
> I'd like to take a couple of hours to make sure the Dispatcher lifetimes definition is not too restrictive for some corner cases.

I'm not rushing you btw. By "soonish" I meant something like "sometime this week", not "absolutely today". ^^"

fengalin:
> By "soonish" I meant something like "sometime this week", not "absolutely today". ^^"

:) np! I wanted to handle this anyway otherwise I'd have kept on procrastinating.

I just pushed a commit for a proposal regarding the `Token` thing. The idea is to keep a single `Token` type, remove the `Copy` / `Clone` implementations, use references in function which shouldn't consume the `Token` and consume it in functions after which the `Token` should no longer be used.

One thing that comes to mind: for `Source`s that need to operate on the `token.sub_id`, we might want dedicated functions for out of crate implementations. See how it's done currently in https://github.com/fengalin/calloop/blob/d11905df9aa32d2725ba9e13f21f9da8788ef005/src/loop_logic.rs#L552.

vberger:
Yes, the current way it is done by destructuring can only work in the crate, due to field visibility. The thing is, creating a new token with a modified subfield basically amounts to cloning the token & changing the `sub_id` field.

So, for implementations that need to touch the `sub_id`, the token needs to be cloneable. And we don't want the token traking registrations to be cloneable. I don't see how wecan meet these requirements without two distinct types...

fengalin:
> I don't see how we can meet these requirements without two distinct types...

See my last proposal in https://github.com/Smithay/calloop/pull/20/commits/2892e15942ea64a3c130fe66f06cf739f1b0444e.

vberger:
Okay, I see indeed this may works as far as I can tell.

Just, I'm having a hard time to get the point, what is problematic about just having two different types?

fengalin:
> Just, I'm having a hard time to get the point, what is problematic about just having two different types?

I though it was elegant, but I don't mind using another type. The best name I can find is `RegistrationToken`. I'll go with that if you agree.

vberger:
I think going with two types would be the simplest for users of calloop, this way it'd be explicit that there are two distinct uses that don't overlap.

`RegistrationToken` for the type used in the `EventSource` trait you mean? Sounds good to me.

fengalin:
I think it's ready for review now.

src/loop_logic.rs
fengalin: Indeed. I removed the comment.

CHANGELOG.md
fengalin: Added some details. Hope this better now.

vberger (approved):
Okay, this is good, many thanks!

PR #28, opened 2020-10-06
channel: signal readinnes after sending message

vberger:
Doing it the other way around can be racy in multithreaded contexts,
causing the event loop to wake-up and try to read the message before it
has even been written, and thus missing it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+#### Fixes
+
+- Channel now signals readinnes after the event has actually been sent, fixing a race
+  condition where the event loop would try to read the message before it has been
+  written.
+
 ## 0.6.4 -- 2020-08-30
 
 #### Fixes
--- a/src/sources/channel.rs
+++ b/src/sources/channel.rs
@@ -46,8 +46,7 @@ impl<T> Sender<T> {
     /// This will wake the event loop and deliver an `Event::Msg` to
     /// it containing the provided value.
     pub fn send(&self, t: T) -> Result<(), mpsc::SendError<T>> {
-        self.ping.ping();
-        self.sender.send(t)
+        self.sender.send(t).map(|()| self.ping.ping())
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/sources/channel.rs
```diff
@@ -85,19 +84,26 @@ impl<T> SyncSender<T> {
     /// deliver the message. Readiness is signalled to the event loop
     /// *before* blocking.
     pub fn send(&self, t: T) -> Result<(), mpsc::SendError<T>> {
-        self.ping.ping();
-        self.sender.send(t)
+        let ret = self.try_send(t);
+        match ret {
+            Ok(()) => Ok(()),
+            Err(mpsc::TrySendError::Full(t)) => self.sender.send(t),
```
kchibisov, line 90: It seems like you don't ping after this `send`?

src/sources/channel.rs
```diff
@@ -85,19 +84,26 @@ impl<T> SyncSender<T> {
     /// deliver the message. Readiness is signalled to the event loop
     /// *before* blocking.
```
kchibisov, line 85: I feel like you should remove the last sentence, since it's talking purely about 'internals', that are not meaningful to the user.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+#### Fixes
+
+- Channel now signals readinnes after the event has actually be sent, fixing a race
```
kchibisov, line 7:
```suggestion
- Channel now signals readinnes after the event has actually been sent, fixing a race
```

kchibisov (changes requested):

src/sources/channel.rs
vberger: The previous call to `self.try_send()` will have pinged already.

src/sources/channel.rs
vberger: This seems like internals, but it is observable behavior, similar to the issue you reported in the chatroom. This guarantees that the event loop will be woken up and thus that the app should not block indefinitely.

kchibisov: >This seems like internals, but it is observable behavior, similar to the issue you reported in the chatroom. This guarantees that the event loop will be woken up and thus that the app should not block indefinitely.

But I'd assume that user expects that adding things to source should wake up a channel? In anyway, you should s/before/after, you really want to keep that.

src/sources/channel.rs
kchibisov: Yes, but there's no guarantee that it's `try_send` and `send` are done atomically? So you can ping() -> poll() -> send() -> lost wake up.

vberger: You mean, if the event loops wakes up and empties the full channel before the sending thread gets the occasion to send its message?

src/sources/channel.rs
vberger: Well, no? If you block before signalling readiness, then the event loop will never wake up, the channel will never be emptied, and you'll never stop blocking... ?

vberger: If we get `Disconnected`, that means that the receiver side of the channel has been dropped, so there is nothing to ping anymore.

src/sources/channel.rs
vberger: I suspect the post-ping would mostly cause spurious wakeups of the event loop overall, but I guess we can err on the side of safety. Spurious wakeups are better than delayed events.

src/sources/channel.rs
kchibisov: I mean that you can't wake up event loop if you've disconnected, so it's fine to ping, since it's 'free'?

vberger: It one syscall, so it's not completely free. So yeah, calling the ping method would work, but there is no point in doing so.

kchibisov (approved):

Issue #36, opened 2021-02-15
Struggling to integrate with socket-based code ⇒ API suggestions


detly:
I've been trying to integrate some (ZeroMQ) socket based code with Calloop, but I'm running up against lifetime errors around the same kind of pattern over and over again. Here's an example (without any ZeroMQ code at all):

```rust
use std::collections::VecDeque;

use calloop::generic::{Fd, Generic};
use calloop::signals::{Signal, Signals};
use calloop::{EventLoop, Interest, LoopHandle, LoopSignal, Mode, Readiness};

fn main() {
    // EventLoop and the handles it produces are parameterised by both a type
    // and lifetime. The lifetime is related to the lifetime of the stored
    // callbacks and sources. The definitions look like:
    //     pub struct EventLoop<'l, Data> {
    //         handle: LoopHandle<'l, Data>,
    //         ...
    let mut event_loop: EventLoop<LoopSignal> = EventLoop::try_new().unwrap();

    // Parameterised with same lifetime 'l as event_loop.
    let handle = event_loop.handle();

    // Exit on interrupt (for Ctrl-C when running in console) and terminate (for
    // OpenWrt's service control).
    const SIGNALS_TO_STOP_ON: &[Signal] = &[Signal::SIGINT, Signal::SIGTERM];

    let handle = event_loop.handle();

    handle
        .insert_source(
            Signals::new(SIGNALS_TO_STOP_ON).unwrap(),
            |event, _metadata, signaller: &mut LoopSignal| {
                println!("Signal received: {}", event.signal());
                signaller.stop();
            },
        )
        .unwrap();

    let mqueue = MessageQueue::new();
    mqueue.register(&handle);

    event_loop
        .run(None, &mut event_loop.get_signal(), |_| {})
        .expect("Could not run event loop");
}

struct Message<'f> {
    message: String,
    callback: Option<Box<dyn FnOnce() + 'f>>,
}

struct MessageQueue<'q> {
    queue: VecDeque<Message<'q>>,
}

impl<'q> MessageQueue<'q> {
    fn new() -> Self {
        MessageQueue {
            queue: VecDeque::new(),
        }
    }

    fn register<'l, D>(&self, handle: &LoopHandle<'l, D>) {
        let event_source = Generic::from_fd(0, Interest::READ, Mode::Edge);

        let callback = |_: Readiness, _: &mut Fd, _: &mut D| {
            self.on_socket_writeable();
            Ok(())
        };

        handle
[...]

PR #37, opened 2021-03-01
Add a method on Token to create a copy with the provided sub_id.

detly:
See discussion on #36. This just means you can do the below when combining event sources:

```rust
fn register(&mut self, poll: &mut Poll, token: Token) -> std::io::Result<()> {
    self.socket.register(poll, token.with_sub_id(SOCKET_READY))?;
    self.incoming_messages.register(poll, token.with_sub_id(CHANNEL_READY))?;
    Ok(())
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+#### Additions
+
+- `Token` now has a method `with_sub_id()` that returns a copy of the token
+  but with the given `sub_id`.
+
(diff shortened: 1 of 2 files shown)


vberger:
Ah, thanks for the PR. This is exactly what I had in mind!

Could you please just add a changelog entry describing this addition?

detly:
Done! Also had to correct the author info because I used the wrong config.

vberger:
Thanks!

PR #40, opened 2021-05-11
Rewrite the front page example to compile and run

detly:
Rewrite the front page example to compile and run, demonstrating using a timer to exit after 5s.

I thought it would be useful for new users if the example program in the README actually compiled and ran. I picked a timer because it has the simplest API and is self contained. Note I've dropped some pre-2018-edition-isms for the sake of space.

I won't be offended if you reject this, it's from example code I had and maybe it's not your goal with a README example. Also, since I added about 30 lines of comments, let me know if you want me to try and cut it back a bit, because it does push other material down the front page.


detly:
> FYI, I'm already using cargo-readme to manually generate the readme if this repo, as attests the README.tpl.

Haha oops. Look, most of the effort was in adding some explanatory comments and making the example itself run. Let me look at getting it to work with `cargo-readme` and I'll add something here soon.

detly:
Right, so running `cargo readme > README.md` does not change the code example. However, it does change some other things, mainly wrapping (meh), but also code links of the form `[`EventLoop`]` etc. These appearing in README.md will still have the square brackets around them because there's no Rustdoc run to interpret them as links. There will also be some other broken links (again, Rustdoc knows where they should point to, but not Github). Let me know if you want me to do anything about this.

vberger:
Sorry to have let this sit for so long, thanks a lot for improving the main example!

Issue #32, opened 2021-01-28
Better docs on how to integrate with futures.


cheako:
I'm not sure if I can even use adapt_io if I decide to use tokio.  It takes a file descriptor, but afaict EventSource don't have anything to do with those.

Likewise it's not clear how to call async functions using calloop, so I feel I'm going to eventually need to get EventSource playing nicely with tokio.

https://gitlab.com/cheako/hidcat/-/commit/86f0b34c1fef0922fea0112e077efc721d33e5f1

vberger:
`adapt_io` is meant to provide a way to wake up powered by calloop, so *instead* of using Tokio or async_std.

Indeed, the docs are missing examples, in the meantime you can see an example of how to use adapt_io and the executor in this test : https://github.com/Smithay/calloop/blob/017e0cf5018d8778656cc6bbb0fbf8d6bde8c152/src/io.rs#L324

vberger:
I believe this is now addressed by the [calloop tutorial](https://smithay.github.io/calloop/)

cheako:
Currently, throwing a 404: https://smithay.github.io/calloop/chp03-00-async-await.html

Note: winit is also an interesting target, for embedded(ala xephyr or xnest) like compositors.

vberger:
> Currently, throwing a 404: https://smithay.github.io/calloop/chp03-00-async-await.html

The broken links are now fixed.

> Note: winit is also an interesting target, for embedded(ala xephyr or xnest) like compositors.

I'm sorry I don't understand what you're suggesting here.

PR #42, opened 2021-06-12
Created mdBook-based documentation with examples

detly:
A while back I wrote #36, and you pointed out that Calloop was designed for composition (rather than whatever I was doing). Your advice there was *extremely* helpful, and led me to make major progress. Thanks!

You also said:

>  The documentation of calloop clearly misses a general overview of "how to do what".

This PR is my effort at writing some documentation that covers that gap. It takes the form of an [mdBook](https://rust-lang.github.io/mdBook/) (same as the Rust official docs) and covers:

- the basics of an event loop
- a simple timer example (same as the README example)
- a more complex ZeroMQ-based example

To allow them both to be hosted on Github pages, I've relegated the API docs to a subdirectory and put the mdBook docs at the root. This means that the mdBook front page acts as a landing page for both the tutorial and the API docs. If you don't like that approach, I'll see if I can reorganise things.

You can it live on my fork's pages: https://detly.github.io/calloop/

I am absolutely open to edits and feedback, no rewrite too big, no detail too small!


detly:
As an aside (ie. not something that should necessarily block this), there could be scope for some extra, more specific, "how to do X in Calloop" chapters. Such as:

- async/futures integration as mentioned in #32 
- how to use Calloop with a raw network socket (probably the async IO adapters? I don't actually know)

If you can provide an answer in the way of some quickly written code examples, I'd be happy to flesh them out into extra documentation. I'd suggest dumping them in separate issues and tagging me.

vberger:
Wow, that is fantastic! Thanks a lot for this effort! :+1: 

I'm sorry I don't have the time to review this in detail right now however, but I should be able to do it during the next week.

detly:
No rush, take your time. I was on leave :wink: I might tweak it as we go, but from this point I won't do any force-pushes or rebases unless necessary.

doc/src/ch02-00-a-simple-example.md
```diff
@@ -0,0 +1,79 @@
...
+## Adding sources to the loop
+We have an event source, we have our shared data, and we know how to start our loop running. All that is left is to learn how to combine these things:
+
+```rust
+{{#rustdoc_include timer_example.rs:insert_source}}
+```
+
+Breaking this down:
+
+- We obtain a handle from the event loop. Much like the timer handle above, this is a separate type from the event loop and allows us to insert sources from event callbacks, other threads, and futures.
```
vberger, line 59: Not exactly, the `LoopHandle` is *not* threadsafe, we cannot use it to manipulate the set of sources from other threads (which would be incredibly racy).

vberger:
Thanks a lot for this tutorial/book, I believe it does a pretty good job at showing how calloop is to be used! :+1:

Given your other issues, it appears we are going to do some small be breaking changes to calloop's API. How do you want to handle that? Are you ok with updating this PR with those changes before we merge it and release the next breaking version of calloop?

vberger:
> If you can provide an answer in the way of some quickly written code examples, I'd be happy to flesh them out into extra documentation. I'd suggest dumping them in separate issues and tagging me.

BTW, regarding futures and async_io handling, there are some code examples in calloop's test suite, is that enough for you?

https://github.com/Smithay/calloop/blob/0f906b24cd0cb07081ddb351a792617d791709cc/src/io.rs#L314-L367

doc/src/ch02-00-a-simple-example.md
detly: Good catch.

detly:
> Are you ok with updating this PR with those changes before we merge it and release the next breaking version of calloop?

Yes, I expected exactly this would happen once I tried to write out some more complex sources! Go ahead and I'll make the edits.

detly:
> BTW, regarding futures and async_io handling, there are some code examples in calloop's test suite, is that enough for you?

Yes, that'll do nicely.

detly:
Feel free to start checking through the text and code examples in chapter 3. I don't plan for that to change now except for corrections. All that's left for me is to add some examples in chapter 4 and then I'll be done.

doc/src/ch03-04-creating-our-source-part-3-processing-events-almost.md
```diff
@@ -0,0 +1,161 @@
...
+    F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
+```
+
+What a mouthful! But when you break it down, it's not so complicated:
+
+- We take our own state, of course, as `&mut self`.
+
+- We take a `Readiness` value - this is mainly useful for "real" file descriptors, and tells you whether the event source was woken up for a read or write event. We ignore it though, because our internal sources are always only readable (remember that even if the zsocket is writeable, the FD it exposes is only ever readable).
+
+- We take a token, just like our register/re-register methods. The ID will always correspond to our own source, but we can check the sub-ID to see which of our internal sources caused us to run.
```
vberger, line 22: The `sub_id` is no longer visible, you updated the example accordingly, but this sentence is now outdated.

doc/src/SUMMARY.md
```diff
@@ -0,0 +1,20 @@
...
+- [How an event loop works](ch01-00-how-an-event-loop-works.md)
+- [A simple example](ch02-00-a-simple-example.md)
+- [A full example: ZeroMQ](ch03-00-a-full-example-zeromq.md)
+  - [Composition](ch03-01-composition.md)
+  - [Creating our source, part I: our types](ch03-02-creating-our-source-part-1-our-types.md)
+  - [Creating our source, part II: setup methods](ch03-03-creating-our-source-part-2-setup-methods.md)
+  - [Creating our source, part III: processing events (almost)](ch03-04-creating-our-source-part-3-processing-events-almost.md)
+  - [Creating our source, part IV: processing events (really)](ch03-05-creating-our-source-part-4-processing-events-really.md)
+  - [The full ZeroMQ event source code](ch03-06-the-full-zeromq-event-source-code.md)
+- [How to do X in Calloop](ch04-00-how-to-do-x-in-calloop.md)
```
vberger, line 14: As a general organization, it feels to me that "How to do X" should come before "How to write my custom event source" in the chapter list.

vberger (commented):
I have a general comment regarding what you've started writing in the `async` chapters. One of the selling points of calloop is that it is completely single-threaded: the IO and general powering of the futures is handled by the `EventLoop` itself.

Using `async_std` kind of defeats this purpose, as it spawns its own logic in a new thread to monitor file descriptors and timers and wake up the futures. However that functionality is not (yet?) provided by calloop, as calloop is at its core not really centered on rust's `async`.

detly:
> Using async_std kind of defeats this purpose, as it spawns its own logic in a new thread to monitor file descriptors and timers and wake up the futures.

Oof, I did not realise this. I was aiming for a simple way to show how to combine Calloop with async code, especially because (a) there are a fair amount of async Rust examples already floating around, and this gives people a way to "transition" async code to Calloop, and (b) there are things that can't easily be accomplished without async libs such as using subprocesses. In fact, in the "X = subprocesses" section basically says "use `async_process`, which also uses a background thread to accomplish asynchronous subprocess management. (I don't know of an event loop library in any language that *doesn't* do this though.)

I wanted to avoid combining *two* new concepts in the same how-to section, and leave the `adapt_io()` feature for a different section. Is this folly? Is there another way to demonstrate running an async block in Calloop without the extra first step of converting blocking IO types first?

doc/src/ch03-04-creating-our-source-part-3-processing-events-almost.md
detly: Good catch, thanks.

vberger:
> I wanted to avoid combining two new concepts in the same how-to section, and leave the adapt_io() feature for a different section. Is this folly? Is there another way to demonstrate running an async block in Calloop without the extra first step of converting blocking IO types first?

Well, this might need some more adapters in calloop. A way to achieve this with current calloop would be to use an [async channel](https://docs.rs/futures/0.3.15/futures/channel/index.html) that would be fed by an other source (a timer for example), but we may consider leaving these parts of the book for later, once the adapters are indeed implemented in calloop.

detly:
> A way to achieve this with current calloop would be to use an async channel that would be fed by an other source (a timer for example)

That's a good idea, actually. I was fixated on timers because the usual examples use `sleep()`, but there's no reason it can't be a channel or other tool in `futures::io`.

Or, as you say, I could leave that part for another time.

detly:
I just updated the async example, let me know what you think.

doc/src/SUMMARY.md
detly: I'd be happy with that, but I'll do at it the end of the edits to keep the diff relatively clean.

doc/src/async_example.rs
```diff
@@ -0,0 +1,76 @@
...
+    // Just like in the timer example, we want to use a LoopSignal as the loop
+    // data so we can exit the program later.
+    let mut event_loop: EventLoop<LoopSignal> = EventLoop::try_new()?;
+    let handle = event_loop.handle();
+
+    handle.insert_source(exec, |evt, _metadata, loop_signaller| {
+        // Print the value of the async block ie. the return value.
+        println!("Async block ended with: {}", evt);
+        // Stop the loop when all blocks have run.
+        loop_signaller.stop();
```
vberger, line 23: Doing so will actually stop the loop as soon as *any* future returns a value.

detly: > Doing so will actually stop the loop as soon as any future returns a value.

Oops. Begs the question of why I get the output that I do, but perhaps it's a timing thing. I'll push a fixed version.

detly:
Just added an example for adapting blocking IO types with `LoopHandle::adapt_io()`.

vberger:
> Just added an example for adapting blocking IO types with `LoopHandle::adapt_io()`.

Thanks, that looks pretty nice! A small caveat I'd have would be that this section is really about adapting IO types for uses with `async`, rather than just using IO type in calloop (which is mostly handled by `Generic`). Maybe it'd appropriate to split the "how to do X" section in two, on the one side a presentation of the few event sources already provided with calloop, and on the other hand a section about "how to integrate calloop with `async` code".

What do you think?

detly:
Thanks for the kind words!

> Maybe it'd appropriate to split the "how to do X" section in two, on the one side a presentation of the few event sources already provided with calloop, and on the other hand a section about "how to integrate calloop with async code".

This really depends on how people tend to *seek* the information in the first place, which you would be better placed to answer than me I think. A lot of documentation is written in a top-down, theory-first kind of way when a lot of users often just want a "how do I get up and running with X"; at the other end of the spectrum, some docs are just "here's a bin of a examples, sift through them if you like" which can also be frustrating. So it depends on what kind of questions you find yourself fielding more often.

I had issue #32 in mind when I started the "how to do X" session. As far as *"a presentation of the few event sources already provided with calloop"*, I feel like the ZeroMQ example shows how to use things like channels, pings, etc, but maybe they also need to be covered out of that context to be more accessible..?

Noted your comment about the IO adapter comments, I'll add something about that to clarify.

vberger:
Mostly, my concern is that, while calloop does provide adapters to integrate with the async ecosystem, it has not been designed as an async-first event loop like tokio or async-std. As such, I do now want in the documentation to present async first as the default choice for event handling in calloop, because it isn't.

As it is now, the "How to do X" chapter in your PR is exclusively about async-related constructs, this makes a stark contrast with the "A simple example" chapter which is not about async at all.

Coming from this, I believe a structure like the following may be well aligned with calloop's general spirit:

1. "How an event loop works" / general presentation of the main calloop principles
2. "Using event sources" / presentation akin the current "a simple example" of how the event sources are used, with a few more examples, ideally one example per provided event source
3. "But what if some parts of my logic are better expressed with async/await?" / presentation of the executor and adapt_io mechanisms, that allow to mix async futures into the your calloop logic. For example to do stuff like fetch a remote file with async/await and once it is retrieved & parsed, forward it back to the main event loop via a channel. Things like that.
4. "Building your own event source: a detailed example" / the detailed ZeroMQ tutorial you wrote already

It's not very different with the current state of your PR, it's mostly a matter of renaming some sections & rewording a few things. So this does not necessarily needs to be done as this PR (as said, its is already very good!). I can take care of rewording and expanding parts 2 and 3 myself after merging your PR if you've had enough with all this re-organizing & such.

detly:
Okay, I see what you mean. Yeah, I think the weird async focus is just a quirk of the fact that it's the last stuff I learnt — the simpler stuff are things I now just check the API docs for when I need to remember something. That said, maybe some of what is now chapter 3 should simply... go in the API docs? In the module level documentation, I mean.

I'd be perfectly happy for you to merge it in now and tinker with it going forward.

vberger:
> That said, maybe some of what is now chapter 3 should simply... go in the API docs? In the module level documentation, I mean.

That is a good point, it's true that the API docs of this part are still very minimalistic. But overall I think it's good to have some redundancy between the API docs and the tutorial, so it does not necessarily need to be removed from the mdbook.

vberger:
FYI I pushed a commit on master doing the core of the remaining reorganization I had in mind. I'll continue to flesh-out ch02 during the weekend I think.

detly:
I agree re. redundancy. It certainly couldn't hurt. I had another thought actually - maybe what I called "how to do X in Calloop" should simply be its own short chapter on "how to mix Calloop and regular async code". And then there can simply be an earlier chapter on those basic elements and how to use them.

vberger:
That's mostly how I started reorganizing things: https://smithay.github.io/calloop

PR #46, opened 2021-06-14
Allow EventSource to request a post-dispatch Action

### LINKED ISSUE #45: Is there a way for a source to remove itself within its own process_events()?
I have recently found the need to have event sources remove themselves from the loop after they complete a task. Mostly this is with timers and subprocesses (via futures), but it can be demonstrated quite simply with an example. Here's an event source that starts a timer when it's added to the loop, and when the timer expires, prints a message and never runs again.

```rust
use calloop::{EventLoop, EventSource, Poll, Readiness, Token};

struct ThingOne {
    timer: calloop::timer::Timer<()>,
    done: bool,
}

impl ThingOne {
    fn new() -> Self {
        Self {
            timer: calloop::timer::Timer::new().unwrap(),
            done: false,
        }
    }
}

impl EventSource for ThingOne {
    type Event = ();
    type Metadata = ();
    type Ret = ();

    fn process_events<F>(&mut self, readiness: Readiness, token: Token, _: F) -> std::io::Result<()>
    where
        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
    {
        if !self.done {
            let done = &mut self.done;
            self.timer.process_events(readiness, token, |_, _| {
                println!("Done!");
                *done = true;
            })?;
        }

        Ok(())
    }

    fn register(&mut self, poll: &mut Poll, token: Token) -> std::io::Result<()> {
        let hdl = self.timer.handle();
        hdl.add_timeout(std::time::Duration::from_secs(1), ());
        self.timer.register(poll, token)
    }

    fn reregister(&mut self, poll: &mut Poll, token: Token) -> std::io::Result<()> {
        let hdl = self.timer.handle();
        hdl.add_timeout(std::time::Duration::from_secs(1), ());
        self.timer.reregister(poll, token)
    }

    fn unregister(&mut self, poll: &mut Poll) -> std::io::Result<()> {
        self.timer.unregister(poll)
    }
}

fn main() {
    let mut event_loop = EventLoop::try_new().unwrap();

    let _loop_token = handle.insert_source(ThingOne::new(), |_, _, _| {}).unwrap();

    event_loop
        .run(None, &mut (), |_| {})
        .expect("Error during event loop!");
}
```

It would be good if I could actually have `ThingOne` remove itself from the loop when it's done. In this case, it's not a big deal. But when you're dealing with open file descriptors and sockets, it's good to have cleanup. But I can't figure out how.

- I can't remove it from inside the `process_events()` method, because I don't have the registration token or the poll argument for `self.unregister()`.
- I can't remove it via `main()`, because by the time I have the registration token, the loop has taken ownership of the source.

Is there a way to do this that I'm missing?

vberger:
Fixes #45


vberger:
cc @detly, does this API meet your needs?

src/sources/mod.rs
```diff
@@ -15,6 +15,23 @@ pub mod ping;
...
+/// Possible actions that can be requested to the event loop by an
+/// event source once its events have been processed
+pub enum PostAction {
+    /// Do nothing specific
+    None,
+    /// Trigger a re-registration of this source
+    Reregister,
+    /// Disable this source
+    ///
+    /// Has the same effect as [`LoopHandle::enable`](crate::LoopHandle#method.enable)
```
detly, line 27: Should this be `LoopHandle::disable`?

src/sources/mod.rs
```diff
@@ -15,6 +15,23 @@ pub mod ping;
 pub mod signals;
 pub mod timer;
 
+/// Possible actions that can be requested to the event loop by an
+/// event source once its events have been processed
+pub enum PostAction {
+    /// Do nothing specific
+    None,
```
detly, line 22: It might be confusing to name this `None`, being identical to one of the `Option` variants. Perhaps... `Continue` or `Await` or something?

detly:
Is `unregister()` called on `PostAction::Remove`?

src/sources/mod.rs
vberger: yes indeed :sweat_smile:

vberger:
> Is `unregister()` called on `PostAction::Remove`?

Yes, in the `dispatch_events` method, in the code block after the `match`

PR #48, opened 2021-06-14
Introduce TokenFactory for unique token generation

### LINKED ISSUE #44: Token sub-ID system does not allow more than one level of composition
Consider two custom event sources:

```rust
struct SourceOne {
    ping_one_a: calloop::ping::PingSource,
    ping_one_b: calloop::ping::PingSource,
}

struct SourceTwo {
    ping_two_a: calloop::ping::PingSource,
    ping_two_b: calloop::ping::PingSource,
}
```

Following the composition technique suggested in #36, you might create either of these with their own token sub-IDs:

```rust
impl SourceOne {
    const ID_A: u32 = 1;
    const ID_B: u32 = 2;
}

impl SourceTwo {
    const ID_A: u32 = 1;
    const ID_B: u32 = 2;
}
```

This works fine, you can use the `ID_*` sub-IDs to differentiate the ping sources in either custom event source.

Now consider what happens if you want to create an event source that uses either or both of those. Where should an API user start *their* token sub-IDs?

I am making the assumption here that any token sub-IDs used by a custom source are implementation details. If that's the case, then a writer of a new source `SourceThree` that contains `SourceOne` plus another source would not know to start their sub-IDs at 3 instead of 1. If they don't, then they will collide: they only get the one `token` in `register()`, and so in `process_events()` there is no way to distinguish eg. `SourceOne::ping_one_b` from whichever `SourceThree` source has sub-ID 2.

Even if we don't make that assumption, there's still a problem. If `SourceThree` wants to combine *both* `SourceOne` and `SourceTwo`, the sub-IDs will always collide, and `SourceThree::process_events()` will have no way to distinguish whether a given event was due to `SourceOne::ping_one_a` or `SourceTwo::ping_two_a`. So even if sub-IDs are considered part of the API of an event source, any given event source needs to know about every other kind of event source, *even any created and published by 3rd parties*, to avoid collisions.

Or have I missed something?

vberger:
This changes the sub_id token assignation mechanism to be dynamical,
rather than static. Allowing multiple-level nesting of event sources,
each one now being responsible to decide if it needs to run or not.

Fixes #44


vberger:
cc @detly

src/sources/generic.rs
```diff
@@ -96,25 +98,44 @@ impl<F: AsRawFd> EventSource for Generic<F> {
...
 
-    fn register(&mut self, poll: &mut Poll, token: Token) -> std::io::Result<()> {
-        poll.register(self.file.as_raw_fd(), self.interest, self.mode, token)
+    fn register(
+        &mut self,
+        poll: &mut Poll,
+        token_factory: &mut TokenFactory,
+    ) -> std::io::Result<()> {
+        let token = token_factory.token();
+        poll.register(self.file.as_raw_fd(), self.interest, self.mode, token)?;
```
detly, line 119: I haven't fully thought this through, but maybe `[re]register()` should return the token on success?  Then you could do `self.token = poll.register(...)` which might be a bit more ergonomic. Of course, that's entirely a matter of opinion, some may prefer it to be more explicit. (Plus you'd still need to return `self.token` then.) Think about it.

detly:
Another thought: what if you used `Option<Token>` in the higher level event sources, and then `None` would be the invalid value. If your `Token` used that trait for "[this number is never zero](https://doc.rust-lang.org/core/num/index.html)" then you'd even get the [zero sized option optimisation](https://doc.rust-lang.org/std/option/index.html#representation) for free. Users could do things like `if self.token.[contains](https://doc.rust-lang.org/std/option/enum.Option.html#method.contains)(passed_in_token)` (if they're on nightly; same for [insert](https://doc.rust-lang.org/std/option/enum.Option.html#method.insert)).

Having said all that... I'm not sure how you'd juggle this, because some parts of the API would want to assume a valid token, and others would not. Semantically it seems right, but in terms of library usage I don't know how much of a pain it would be.

src/sources/generic.rs
vberger: Well, given this method is invoked exactly once in `calloop`'s codebase (all other sources build upon `Generic`), and I expect it to be rarely directly invoked outside of it as well (again, most would just compose using the `Generic` source), I'm thinking I'd prefer to keep it like that, without trying to be smart with the design.

vberger:
Again, I'm not quite sure what would be gained through that? With that `TokenFactory`, and assuming most sources would be built on top of `Generic`, a significant portion of calloop's users will never manipulate an `Token` directly, so that seems like trying to optimize for a marginal (if not non-existing) use.

src/sources/generic.rs
detly: Oh I see what you mean — even if I'm composing a source, I don't actually need my *own* token for the *composed* type, do I? This approach means I just recursively pass the token factory in to my sub-sources. Correct?

vberger: Yes, exactly. And similarly, you just invoke `process_events` recursively unconditionnaly, and it is the inner source that compares with its tokens and decides if it actually has something to do.

detly:
> a significant portion of calloop's users will never manipulate an `Token` directly, so that seems like trying to optimize for a marginal (if not non-existing) use

Yes indeed, there's not much case for it now that I better understand how this is used.

vberger:
Do you have other concerns regarding this PR? I believe once it is merged, all the points you raised about calloop's API will be addressed.

detly:
Nope, nothing more!

Issue #57, opened 2021-07-12
Use own error type for API


detly:
One of the little-but-common annoyances I run into with Calloop is that eg. `process_events()` returns a `std::io::Result`, but `std::io::Error` is just about the only Rust error type in existence that doesn't have a single-argument constructor. So I *always* have to tack on `.map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))`. This is compounded by the fact that you can use any type you like for `EventSource`'s associated `Ret`, which is usually going to be a `Result<T, E> where E: std::error::Error` rather than `std::io::Error` (precisely because `std::io::Error` is a pain to use in your own API and arguably is not intended to be).

Should Calloop have its own error (and corresponding result) for `process_events()` (and possibly other trait methods on `EventSource`? It would be possible to have `From` implementations to ensure that `?` still works. Then (a) using `?` on `std::io::Result`s from fd operations would still be fine, and (b) so would using `?` on `core::result::Result` or `anyhow::Result` or errors from `thiserror` etc.

I would be happy to make such a change, but it is potentially a breaking one so I wanted to see if it's something you'd be open to.

vberger:
Yes, definitely. Notably, `process_events()` should return a result that indicates, when an error occurred, if it came from the polling machinery itself or from one of the event sources, in which case it could be a `Box<dyn Error>`. Probably worth thinking to some mechanism that allows to easily detect which source errored, to act upon that if possible/necessary.

detly:
Excellent, we can discuss the general approach here and then I'll code it up.

I tend to avoid `Box<dyn Error>` - basically it doesn't do a whole lot that [`anyhow`](https://docs.rs/anyhow/1.0.42/anyhow/) or [`thiserror`](https://docs.rs/thiserror/1.0.26/thiserror/index.html) don't achieve between them. Both of those libs will happily work with (ie. wrap, allow unwrapping, combine) errors that implement `std::error::Error`, so we should be fine there.

My first suggestion would be to have a `calloop::CalloopError` that
- implements `std::error::Error`
- has `From` impls from `std::io::Error` into itself
  - possibly has a few distinct types that correspond to obvious `std::io::ErrorKind`s without going overboard
- has `From` impls from `std::error::Error` into itself
- has enough additional information that the event source can at least be identified (access it directly might have ownership issues, because the loop owns the source, but eg. the handle's registration token or the source's token could be recorded)

[...]

PR #60, opened 2021-08-14
Use a SlotMap instead of SourcesList

vberger:


vberger:
@detly What do yo think of the 3rd commit of this PR, wrt to #57  ?

detly:
Sorry, I actually made a fair amount of progress on this but then got derailed by a work deadline. I'll push the remainder of my changes in a couple of hours. They're not exactly API compatible with this, but they're not far off.

The only bit I got really stuck on was the mechanics of handing back an event source if it can't be added to the loop. I think it's just a matter of manually implementing a couple of trails though.

vberger:
OK, I'll just merge that without the 3rd commit then.

PR #62, opened 2021-08-30
Custom error type for Calloop

detly:
This is a final-ish draft of creating an error type for Calloop that covers all the major use cases (internal and external). The two things holding it back from not being a draft are:

- I don't yet know how to exclude necessary parts from tarpaulin
- I'd like to use it with a couple of my work projects, but I need to port various parts for Calloop's recent changes

Sorry it took so long, it was hard to accommodate certain niche cases. I think I've made it as ergonomic as possible.

The basic gist is:

- There's a main error type `calloop::errors::Error` which can take conversions from `std::io::Error` and anything that converts into `Box<dyn Error + Sync + Send>`.
- There's a separate error for `InsertError`, because it's a bit special but rare.
- User callbacks or event source code should just be able to use `?` and not worry about mapping to and from IO errors etc.

I've used `thiserror` where possible to remove some boilerplate, it's worked quite nicely but is the source of the tarpaulin issue.


vberger:
So, overall your design looks pretty good, I'll need to read more carefully all the details though.

In the meantime, it appears there are import errors in the `src/sys/kqueue.rs` file, causing all the freebsd tests to fail, and to exclude code from tarpaulin instrumentation, you can use the `#[cfg(not(tarpaulin_include))]` annotation on items or blocks that should be ignored.

detly:
I just realised that there are some typing issues that still prevent this from working with `?` after all. Give me some time to fix that up.

I thought I fixed the `kqueue` code, I must have broken it in a recent rebase. Cheers.

detly:
Closing this PR while I redo some of it.

PR #64, opened 2021-10-16
macOS port

i509VCB:
~~All tests pass except for `loop_logic::tests::change_interests` which currently hangs forever because I have not found look into macOS missing `SOCK_NONBLOCK` yet.~~


i509VCB:
GUI events on mac os involve an entirely proprietary display server I believe.

vberger:
Thanks!

PR #76, opened 2022-02-21
Prevent dangling pointers being left in the system's polling mechanim and reduce the amount of unsafe code in the library

### LINKED ISSUE #74: Possible soundness improvements for recent slotmap changes
I noticed the changes to use slotmap a while back, and the introduction of a few unsafe code blocks and functions. What mainly caught my eye were a couple of safety notices eg.

> # Safety
> You must ensure that the `*const Token` pointer provided remains alive until your event source is re-registered or unregistered.

(I'll stress that what follows is my analysis, which could be mistaken, so if you're a user take it with a grain of salt until a maintainer of the project provides their feedback.)

The functions in question are marked unsafe, and so there's a "contract" with the caller that they Not Do The Thing you warn against. However, it's not a difficult thing to do, especially when using transient sources (I've certainly done it by mistake myself during development). If it does happen, it's a pretty dire consequence - `epoll()` now has callback data that's effectively a dangling pointer into arbitrary memory.

I think there's a way to change the nature of this, so that instead of a dangling pointer, the result is a leak of a file descriptor and token. Still sub-optimal, but not as bad and technically safe by Rust's contract. Basically it follows the "leak amplification" pattern for other unsafe code in the standard library. An upshot is that it also "pushes back" the unsafe boundary a bit further into the internals of Calloop.

It still uses the same token mechanics, but the initial, invalid token is boxed and converted to a raw pointer as early as possible (see `generic.rs`). The `Poll` type is now responsible for creating a new, valid token at the time of registration, and the source that uses it (probably `Generic`) is responsible for freeing/dropping it. This means that if it is not freed, it leaks, instead of the `Generic` freeing the Box when it's dropped but leaving the pointer in the event loop.

You can take a look at my changes in my [`slotmap-tweaks`](https://github.com/detly/calloop/tree/slotmap-tweaks) branch. They are definitely not finished (as in, they don't even compile yet), because I haven't quite sorted out the type changes required to avoid a leak in the `unregister()` process. If you agree with my analysis, I can open an initial PR and you can review it more fully. If not, I'm happy to discuss it, since I'm certainly not an expert in Rust soundness and safety.

detly:
Previously, dropping a source without unregistering it risked leaving a
dangling pointer to token memory in the underlying polling mechanism. Now, the
Poller type is responsible for keeping track of what data the system's poller
(eg. epoll) is still pointing to, and maintaining ownership of it.

This commit also slightly reduces the amount of unsafe code related to pointer
dereferences and adds some notes justifying the safety of these operations.

## Notes not in the commit message

The "reduce amount of unsafe code" comment simply means I took a couple of things like:

    unsafe { *(event.data() as usize as *const Token) }

...and moved the part that can be done without `unsafe` out. It's not a big deal, but IME this means you're less likely to accidentally put _another_ unsafe operation in there without noticing. I also added some comments so that anyone who wants to eg. port or make changes has more information.

### Token changes

How committed were you to the "invalid Token" mechanic? Because I replaced that with `Option<Token>`. I know the comment said that it's to prevent having `Option<Token>` all over the place, but with the recent changes (slotmap + token factory), there's really only a couple of places it occurs now. I can revert only those changes if you like, they're not essential, they just helped while I was hacking and I left them in.

### Leak amplification and panics

If an event source is dropped before being unregistered, the `Token` and file descriptor won't be removed from the map in `Poller`. This means the `Token` memory will never be deallocated. More severely, if the file descriptor is re-used, this will trigger the [panic here](https://github.com/detly/calloop/blob/slotmap-tweaks/src/sys/mod.rs#L237). It's debatable whether this needs to be so strict.

### New dependency

There's a new dependency on `vec_map`.

### Size changes

My changes seem to add about 5kB to the release + stripped rlib on Linux. I'll see if I can pare that down at all.

Fixes #74.

### DIFF
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -236,7 +236,7 @@ impl Poll {
             if self.tokens.insert(index, token_ptr).is_some() {
                 // If there is already a file descriptor associated with a
                 // token, then replacing that entry will leak the token, but
-                // converting it back into an Rc might leave a dangling pointer
+                // converting it back into a Box might leave a dangling pointer
                 // somewhere. We can theoretically continue safely by choosing
                 // to leak, but one of our assumptions is no longer valid, so
                 // panic.
--- a/src/sys/mod.rs
+++ b/src/sys/mod.rs
@@ -1,4 +1,5 @@
-use std::os::unix::io::RawFd;
+use std::{convert::TryInto, os::unix::io::RawFd};
+use vec_map::VecMap;
 
 use crate::loop_logic::CalloopKey;
 
@@ -23,7 +24,6 @@ mod kqueue;
     target_os = "macos"
 ))]
 use kqueue::Kqueue as Poller;
-use slotmap::Key;
 
 /// Possible modes for registering a file descriptor
 #[derive(Copy, Clone, Debug)]
@@ -149,25 +149,6 @@ pub struct Token {
     pub(crate) sub_id: u32,
 }
 
-#[allow(dead_code)]
-impl Token {
-    /// Produces an invalid Token, that is not recognized by the
-    /// event loop.
-    ///
-    /// Can be used as a placeholder to avoid `Option<Token>`
-    pub fn invalid() -> Token {
-        Token {
-            key: CalloopKey::null(),
-            sub_id: std::u32::MAX,
-        }
-    }
-
-    /// Check if this token is the invalid token
-    pub fn is_invalid(self) -> bool {
-        self.key.is_null()
-    }
-}
-
 /// The polling system
 ///
 /// This type represents the polling system of calloop, on which you
@@ -180,6 +161,21 @@ impl Token {
 /// source and delegate the implementations to it.
 pub struct Poll {
     poller: Poller,
+
+    // It is essential for safe use of this type that the pointers passed in to
+    // the underlying poller API are properly managed. Each time an event source
+    // is registered, the token it passes in is Boxed and converted to a raw
+    // pointer to be passed to the polling system by FFI. This pointer is what's
+    // stored in the map. When the event source is re- or unregistered, the same
+    // raw pointer can then be converted back into the Box and dropped, safely
+    // deallocating it. To put it another way, we effectively "own" the Token
+    // memory on behalf of the underlying polling mechanism.
+    //
+    // All the platforms we currently support follow the rule that file
+    // descriptors must be "small", positive integers. This means we can use a
+    // VecMap which has that exact constraint for its keys. If that ever
+    // changes, this will need to be changed to a different structure.
+    tokens: VecMap<*mut Token>,
 }
 
 impl std::fmt::Debug for Poll {
@@ -192,6 +188,7 @@ impl Poll {
     pub(crate) fn new() -> crate::Result<Poll> {
         Ok(Poll {
             poller: Poller::new()?,
+            tokens: VecMap::new(),
         })
     }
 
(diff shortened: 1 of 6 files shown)


detly:
> My changes seem to add about 5kB to the release + stripped rlib on Linux. I'll see if I can pare that down at all.

I couldn't figure this out. I had assumed it was due to the extra code in `src/sys/mod.rs`, but looking at the assembly it's not. If it's important to you, I can keep looking, otherwise I won't bother. I can tell you that on all of my actual binary projects, it hasn't affected binary size.

On another note, I don't understand why the FreeBSD tests are failing - is it the final `grcov` pass?

```none
grcov . --binary-path ./target/debug -s . -t lcov --branch --llvm --ignore-not-existing --keep-only "src/sys/*" --excl-br-start "mod tests \{" --excl-start "mod tests \{" --excl-br-line "#\[derive\(" --excl-line "#\[derive\(" -o lcov.info
06:38:55 �[0m�[31m[ERROR] �[0mA panic occurred at src/producer.rs:544: No input files found
```

src/sys/mod.rs
```diff
@@ -180,6 +161,21 @@ impl Token {
...
+    // stored in the map. When the event source is re- or unregistered, the same
+    // raw pointer can then be converted back into the Box and dropped, safely
+    // deallocating it. To put it another way, we effectively "own" the Token
+    // memory on behalf of the underlying polling mechanism.
+    //
+    // All the platforms we currently support follow the rule that file
+    // descriptors must be "small", positive integers. This means we can use a
+    // VecMap which has that exact constraint for its keys. If that ever
+    // changes, this will need to be changed to a different structure.
+    tokens: VecMap<*mut Token>,
```
vberger, line 178: Why `VecMap<*mut Token>` rather than `VecMap<Box<Token>>`, given the `VecMap` is effectively what owns the tokens?

src/sys/mod.rs
```diff
@@ -209,50 +206,127 @@ impl Poll {
...
+            let token_box = unsafe { Box::from_raw(token_ptr) };
+            std::mem::drop(token_box);
+        } else {
+            // Registration worked, keep the token pointer until it's replaced
+            // or removed.
+            let index = index_from_fd(fd);
+            if self.tokens.insert(index, token_ptr).is_some() {
+                // If there is already a file descriptor associated with a
+                // token, then replacing that entry will leak the token, but
+                // converting it back into an Rc might leave a dangling pointer
```
vberger, line 239:
```suggestion
                // converting it back into a Box might leave a dangling pointer
```

vberger (commented):
Looks pretty good overall :+1: indeed that makes for a much safer API, thanks!

The FreeBSD failure is indeed unrelated to your PR, something must be wrong in the CI script for handling coverage, I'll need to investigate that. Also, I'm not too worried about the binary size point you raise.

src/sys/mod.rs
detly: I thought about this, the issue is that
- turning the `Box` into a pointer consumes the `Box`, so you'd need to either
  - use an `Rc` (which I originally did but it was very verbose)
  - immediately convert the pointer back into a `Box`, store that, give a copy of the pointer to `epoll()`, and then what if you inadvertently drop it in some other way? It just becomes another safety liability.
- the pointer can't deallocate the data by itself, the worst thing it can do is be forgotten about or dropped, which means a leak rather than dangling pointers elsewhere. So you're holding onto something relatively safe, that can only do unsafe things inside an actual `unsafe` block

I think it might be better semantics to think of `epoll()` (or `kqueue()` etc) as owning the heap allocation here, rather than the `Poll` or the `VecMap` member. At least, in between `[re|un]register()` calls. Then it makes sense to keep only the pointer, and to do the `Box<->raw` conversion right at the boundary of the FFI call (almost, anyway).

vberger:
Alright, I like this, thanks a lot! :+1:

PR #79, opened 2022-02-28
Replace a panicking duration calculation 

detly:
…with saturating duration calculation in timerfd event source.

Under certain (intermittent) circumstances, the timerfd scheduler could
calculate a negative duration, which would cause a panic. In the context of
this calculation, the duration could just be clipped to zero, which is what
`saturating_duration_since()` does.

---

I would see this happen about one time in three when running under tarpaulin:

```none
---- loop_logic::tests::non_static_dispatcher stdout ----
thread 'loop_logic::tests::non_static_dispatcher' panicked at 'supplied instant is later than self', library/std/src/time.rs:313:48
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic_display
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:63:5
   3: core::option::expect_failed
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/option.rs:1637:5
   4: core::option::Option<T>::expect
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/option.rs:709:21
   5: std::time::Instant::duration_since
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/time.rs:313:9
   6: calloop::sources::timer::timerfd::TimerScheduler::reschedule
             at ./src/sources/timer/timerfd.rs:36:44
   7: calloop::sources::timer::TimerInner<T>::reschedule
             at ./src/sources/timer/mod.rs:276:13
   8: calloop::sources::timer::TimerInner<T>::insert
             at ./src/sources/timer/mod.rs:231:9
   9: calloop::sources::timer::TimerHandle<T>::add_timeout
             at ./src/sources/timer/mod.rs:117:9
  10: calloop::loop_logic::tests::non_static_dispatcher
             at ./src/loop_logic.rs:848:21
  11: calloop::loop_logic::tests::non_static_dispatcher::{{closure}}
             at ./src/loop_logic.rs:843:5
  12: core::ops::function::FnOnce::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
  13: core::ops::function::FnOnce::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    loop_logic::tests::non_static_dispatcher

test result: FAILED. 25 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.17s

Error: "Test failed during run"
```

Note this, well, note on `duration_since()`:

> Previous rust versions panicked when `earlier` was later than `self`. Currently this method saturates. Future versions may reintroduce the panic in some circumstances. See [Monotonicity].

It's currently inaccurate but still explains the behaviour.

### DIFF
--- a/src/sources/timer/timerfd.rs
+++ b/src/sources/timer/timerfd.rs
@@ -33,7 +33,7 @@ impl TimerScheduler {
 
     pub fn reschedule(&mut self, new_deadline: Instant) {
         let now = Instant::now();
-        let time = TimeSpec::from_duration(new_deadline.duration_since(now));
+        let time = TimeSpec::from_duration(new_deadline.saturating_duration_since(now));
         let time = match self.current_deadline {
             Some(current_deadline) if new_deadline > current_deadline && current_deadline > now => {
                 return;
--- a/src/sources/timer/timerfd.rs
+++ b/src/sources/timer/timerfd.rs
@@ -3,7 +3,7 @@
 use std::os::unix::io::{AsRawFd, RawFd};
 use std::time::Instant;
 
-use nix::sys::time::TimeSpec;
+use nix::sys::time::{TimeSpec, TimeValLike};
 use nix::sys::timerfd::{ClockId, Expiration, TimerFd, TimerFlags, TimerSetTimeFlags};
 
 use crate::generic::Generic;
@@ -42,6 +42,11 @@ impl TimerScheduler {
         };
 
         self.current_deadline = Some(new_deadline);
+
+        // Calling timerfd_settime() (which is what TimerFd::set() effectively
+        // does) with a zero duration will disarm the timer. It should be
+        // handled before this point.
+        assert_ne!(time, TimeSpec::zero());
 
         let expiration = Expiration::OneShot(time);
         let flags = TimerSetTimeFlags::empty();


detly:
Seems to have changed the failure to this:

```none
---- loop_logic::tests::non_static_dispatcher stdout ----
thread 'loop_logic::tests::non_static_dispatcher' panicked at 'assertion failed: fired', src/loop_logic.rs:868:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:48:5
   3: calloop::loop_logic::tests::non_static_dispatcher
             at ./src/loop_logic.rs:868:13
   4: calloop::loop_logic::tests::non_static_dispatcher::{{closure}}
             at ./src/loop_logic.rs:843:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
   6: core::ops::function::FnOnce::call_once
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    loop_logic::tests::non_static_dispatcher

test result: FAILED. 26 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.15s

Error: "Test failed during run"
```

detly:
Right, so the problem is here (note my added prints):

```rust
pub fn reschedule(&mut self, new_deadline: Instant) {
    println!("Rescheduling: {:?}", new_deadline);
    println!("Current deadline: {:?}", self.current_deadline);
    let now = Instant::now();
    let time = TimeSpec::from_duration(new_deadline.saturating_duration_since(now));
    println!("Time (1): {:?}", time);
    let time = match self.current_deadline {
        Some(current_deadline) if new_deadline > current_deadline && current_deadline > now => {
            return;
        }
        _ => time,
    };

    println!("Time (2): {:?}", time);

    self.current_deadline = Some(new_deadline);

    let expiration = Expiration::OneShot(time);
    let flags = TimerSetTimeFlags::empty();
    self.timerfd
        .set(expiration, flags)
        .expect("setting timerfd failed.");
}
```

On a failing run, I get:

```none
Rescheduling: Instant { tv_sec: 33585, tv_nsec: 392670166 }
Current deadline: None
Time (1): TimeSpec(timespec { tv_sec: 0, tv_nsec: 0 })
Time (2): TimeSpec(timespec { tv_sec: 0, tv_nsec: 0 })
thread 'loop_logic::tests::non_static_dispatcher' panicked at 'assertion failed: fired', src/loop_logic.rs:868:13
```

Note the `TimeSpec`. According to `man timerfd_settime()`:

> new_value.it_value specifies the initial expiration of the timer, in seconds and  nanosec‐
> onds.  Setting either field of new_value.it_value to a nonzero value arms the timer.  Set‐
> ting both fields of new_value.it_value to zero disarms the timer.

So my fix disarms the timer instead of panicking. The trouble is, I'm not sure what to do in `reschedule()` to say "this timer has been rescheduled in a way that makes it expire right now." I'll look into it.

detly:
I'd still suggest having this change in here, and perhaps also an `assert_ne!(time, TimeSpec::zero())` before accidentally disarming the timerfd.

The only solution I can think of, since timerfd doesn't support setting a "fire immediately" value, is to use an extra `PingSource` and have something like

```rust
let now = Instant::now();
let time = TimeSpec::from_duration(new_deadline.saturating_duration_since(now));

if time == TimeSpec::zero() {
    self.ping.ping();
    return;
}

let time = match self.current_deadline {
    // ...
```

But this adds complexity and size to the timer struct.

kchibisov:
@detly can't you just write to timerfd descriptor `1`? So you'll wake up event loop right away due to write to it. The `1` means that it was fired ones.

kchibisov:
An alternative solution would be to make sure that timestamp isn't in the past, but adding the resulted time into heap once it was registered in timer source. This will ensure that the time won't be in a past when it comes to rescheduling.

detly:
> can't you just write to timerfd descriptor 1? So you'll wake up event loop right away due to write to it.

I didn't realise it was writeable. But is it a potentially blocking operation?

I don't understand your other comment, sorry. The call stack with the added assert looks like:

```none
   4: calloop::sources::timer::timerfd::TimerScheduler::reschedule
             at ./src/sources/timer/timerfd.rs:55:9
   5: calloop::sources::timer::TimerInner<T>::reschedule
             at ./src/sources/timer/mod.rs:276:13
   6: calloop::sources::timer::TimerInner<T>::insert
             at ./src/sources/timer/mod.rs:231:9
   7: calloop::sources::timer::TimerHandle<T>::add_timeout
             at ./src/sources/timer/mod.rs:117:9
```

If you follow execution through that, I think it's already doing what you're suggesting: in `add_timeout()` it calls `insert(Instant::now() + delay_from_now, data)`, which calls ` self.heap.push(TimeoutData { deadline, ... })` and then does the reschedule.

kchibisov:
>If you follow execution through that, I think it's already doing what you're suggesting: in add_timeout() it calls insert(Instant::now() + delay_from_now, data), which calls self.heap.push(TimeoutData { deadline, ... }) and then does the reschedule. Is that what you meant?

No, I suggest that you reschedule first, return `Instant` from reschedule and use this `Instant` in `self.heap.push`. So the Duration could be pass through and you won't need `Instant::now` anymore.

>I didn't realise it was writeable. But is it a potentially blocking operation?

You can mark it non block during creation of timerfd if it matters, but it shouldn't block.

detly:
> No, I suggest that you reschedule first, return Instant from reschedule and use this Instant in self.heap.push. So the Duration could be pass through to timerfd_settime and you won't need Instant::now anymore. The issue you're initially facing is that the time is already in the past when it reaches reschedule from insert.

I think there's still an edge case here — what if the caller passes in a zero duration (perhaps because of their own saturating calculation)? Should the API prohibit this?

kchibisov:
>I think there's still an edge case here — what if the caller passes in a zero duration

I think it could be wrapped to the next timestamp?

vberger:
Closing in favor of #82

Issue #78, opened 2022-02-26
"Transient" source wrapper (includes channel CPU spike issue)


detly:
I accidentally triggered a 100% CPU spike in my code, and managed to reduce it to this situation:

```rust
use calloop::{
    channel::{channel, Channel},
    signals::{Signal, Signals},
    EventLoop, EventSource, LoopSignal, PostAction,
};

struct HasChannel(Channel<()>);

impl HasChannel {
    fn new() -> Self {
        Self(channel().1)
    }
}

impl EventSource for HasChannel {
    type Event = calloop::channel::Event<()>;
    type Metadata = ();
    type Ret = ();

    fn process_events<F>(
        &mut self,
        readiness: calloop::Readiness,
        token: calloop::Token,
        callback: F,
    ) -> std::io::Result<calloop::PostAction>
    where
        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
    {
        self.0.process_events(readiness, token, callback)?;
        Ok(PostAction::Continue)
    }

    fn register(
        &mut self,
        poll: &mut calloop::Poll,
        token_factory: &mut calloop::TokenFactory,
    ) -> std::io::Result<()> {
        self.0.register(poll, token_factory)
    }

    fn reregister(
        &mut self,
        poll: &mut calloop::Poll,
        token_factory: &mut calloop::TokenFactory,
    ) -> std::io::Result<()> {
        self.0.reregister(poll, token_factory)
    }

    fn unregister(&mut self, poll: &mut calloop::Poll) -> std::io::Result<()> {
        self.0.unregister(poll)
    }
}

fn main() -> std::io::Result<()> {
    // Event loop initialisation.
    let mut event_loop: EventLoop<LoopSignal> = EventLoop::try_new()?;

    const SIGNALS_TO_STOP_ON: &[Signal] = &[Signal::SIGINT, Signal::SIGTERM];

    let handle = event_loop.handle();

    let mut signaller = event_loop.get_signal();

    handle.insert_source(
        Signals::new(SIGNALS_TO_STOP_ON)?,
        |_, _metadata, signaller| signaller.stop(),
    )?;

    handle.insert_source(HasChannel::new(), |_, _, _| {})?;

    event_loop.run(None, &mut signaller, |_| {})?;

    Ok(())
}
```

(This can run as a `main.rs` with only `calloop = "0.9.3"` as a dependency.)

[...]

PR #80, opened 2022-03-02
A transient event source wrapper

detly:
See #78

### DIFF
--- a/src/sources/transient.rs
+++ b/src/sources/transient.rs
@@ -262,3 +262,411 @@ impl<T: crate::EventSource> crate::EventSource for TransientSource<T> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{
+        channel::{channel, Event},
+        ping::{make_ping, PingSource},
+        Dispatcher, EventSource, PostAction,
+    };
+    use std::{
+        sync::atomic::{AtomicBool, Ordering},
+        time::Duration,
+    };
+
+    // This can be removed in favour of Duration::ZERO when the MSRV is 1.53.0
+    const DURATION_ZERO: Duration = Duration::from_nanos(0);
+
+    #[test]
+    fn test_transient_drop() {
+        // A test source that sets a flag when it's dropped.
+        struct TestSource<'a> {
+            dropped: &'a AtomicBool,
+            ping: PingSource,
+        }
+
+        impl<'a> Drop for TestSource<'a> {
+            fn drop(&mut self) {
+                self.dropped.store(true, Ordering::Relaxed)
+            }
+        }
+
+        impl<'a> crate::EventSource for TestSource<'a> {
+            type Event = ();
+            type Metadata = ();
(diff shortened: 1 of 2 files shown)


src/sources/transient.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// The source needs to be registered with the loop.
+    Register(Box<T>),
+    /// The source needs to be disabled but kept.
+    Disable(Box<T>),
+    /// The source needs to be removed from the loop.
+    Remove(Box<T>),
+    /// The source has been removed from the loop and dropped (this might also
+    /// be observed if there is a panic while changing states).
+    None,
+}
```
vberger, line 109: Why the inner `Box`?

vberger (commented):
I'm really liking this wrapper, as I get accustomed to it. I have a feeling that it can be improved further, but that's still not clear to me how exactly, so I'm in favor of merging it for now, I just have one question (see the other comment).

Also, could you rebase your PR on master? I've added a commit that disable coverage on FreeBSD as it is currently broken, so this should fix the CI issues.

src/sources/transient.rs
detly: This was a practical measure to reduce code size, because there's no implementation of `EventSource` for `Box<EventSource>` (or, say, some trait that `Box<EventSource>` would implement, I don't know if that's possible). I was using this in an application that had quite a few `TransientSource<SubprocSource>` where `SubprocSource` had a substantial on-stack size, so the box helped reduce code size and stack pressure in an embedded system. I'd be happy to remove it (it won't kill my application to go back, I was just trying lots of different things), but you'll probably see a PR from me in the near future implementing `EventSource` for `Box<EventSource>` or something like that.

vberger: Oh, I see.

Well, an `impl<T> EventSource for Box<T> where T: EventSource { ... }` seems like a pretty legit thing to add to calloop, yeah. I'm not sure an impl for `Box<dyn EventSource>` would be possible though, because of the associated constants and all.

detly:
> I have a feeling that it can be improved further, but that's still not clear to me how exactly, so I'm in favor of merging it for now

I suspect that with use, any improvements will become clearer. I'll release a couple of crates soon that use it soon, too.

src/sources/transient.rs
detly: Okay, I'll remove the inner boxing and submit a separate PR for the transitive-ness. The associated types shouldn't be a problem, see [Discord chat](https://discord.com/channels/442252698964721669/443150878111694848/949657615233847337).

vberger: Ah yes indeed, looks like this kind of things work:

```rust
trait Foo {
    type Bar;
    
    fn method(&mut self, bar: Self::Bar);
}

impl<B> Foo for Box<dyn Foo<Bar=B>> {
    type Bar = B;
    
    fn method(&mut self, bar: B) {
        (&mut **self).method(bar)
    }
}
```

PR #81, opened 2022-03-04
Add some registration macros to help with composite sources.

detly:
I have these in my own "misc extras" (private) crate and have found them very useful for my top level sources. If you think they'd be a good addition to the core Calloop library, here they are. If not, I'll publish them in an "extras" crate perhaps.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,4 +154,5 @@ pub use error::{Error, InsertError, Result};
 
 pub mod io;
 mod loop_logic;
+mod macros;
 mod sources;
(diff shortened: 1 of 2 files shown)


vberger:
These look like useful helpers for complex composite even sources, and I don't see why not bring then into calloop. However it looks like the CI is pretty unhappy about the doc tests.

Also, I think I'd prefer each macro to have its own documentation example, rather than making the module containing them public.

detly:
> However it looks like the CI is pretty unhappy about the doc tests.

The what. (They're not meant to be doc tests. I shall mark them as not tests, and perhaps they need some real tests too.)

detly:
Separated the docs, made the module non-public and added a test (even though it doesn't seem to affect code coverage?).

vberger:
thanks!

PR #85, opened 2022-03-07
Add passthrough implementations for EventSource for Box<T> and &mut T.

detly:
This allows some control over the memory placement of event sources without the need to change the code that uses the trait.

Note that the `EventSource` trait turns out not to be object-safe, not because of the associated types, but because of the generics on eg. `process_events()`. No matter.

### DIFF
--- a/src/sources/mod.rs
+++ b/src/sources/mod.rs
@@ -144,6 +144,109 @@ pub trait EventSource {
     }
 }
 
+/// Blanket implementation for boxed event sources. [`EventSource`] is not an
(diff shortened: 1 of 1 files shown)


detly:
Gosh, this might be the first time in my entire professional and FOSS-contributor life I've submitted a PR with 100% green merge checks on the first commit.

detly:
Missed the default methods, that's what I get for hubris.

vberger:
Thanks!

PR #89, opened 2022-03-21
Rework timer event sources around Poll

### LINKED ISSUE #9: Use OS-based sources for timer when possible
This is at least possible on linux by using a timerfd. I don't know about other platforms.

This would allow use to only spawn a timer thread when needed.

### LINKED ISSUE #83: Add self-repeating timers
In some contexts, there is use for cases like "start a timer for XX seconds and then repeat it every YY seconds until cancelled" (for example keyboard repeat in Wayland), we could provide an API for that for convenience.

### LINKED ISSUE #84: Investigate non-deterministic timer test failure on macos
The current timer implementation on macos randomly fails the CI in a non-deterministic way. It'd be nice if someone with access to a macos computer could investigate it.

vberger:
The reworks the `Timer` event source to have it directly driven
by the core polling of the event loop, instead of relying on
external timing mechanisms (like a thread or a timerfd).

This PR does not yet update the book chapter about timers in consequence, I'd first like to have your opinion on this change @detly and @kchibisov 

Fixes #9
Fixes #83
Fixes #84


README.md
```diff
@@ -65,16 +55,22 @@ fn main() {
...
                 // The return type is just () because nothing uses it. Some
                 // sources will expect a Result of some kind instead.
-                println!("Event fired: {}", event);
+                println!("Timeout for {:?} expired!", event);
+                // notify the event loop to stop running using the signal in the shared data
+                // (see below)
                 shared_data.stop();
+                // The timer event source requires us to return a TimeoutAction to
+                // specify if the timer should be rescheduled. In our case we just drop it.
+                TimeoutAction::Drop
```
detly, line 73: Nice!

src/sources/timer.rs
```diff
@@ -0,0 +1,475 @@
+//! Timer event source
+//!
+//! The [`Timer`] is an event source that will fire its event after a certain amount of time
+//! specified at creation. Its timing is tracked directly by the event loop core logic, and it does
+//! not consume any system resource.
+//!
+//! The timer precision depends on whever the  was initialized in high-precision mode. If not,
```
detly, line 7:
```suggestion
//! The timer precision depends on whether the loop was initialized in high-precision mode. If not,
```

detly:
I like what I've taken in on the first pass! Especially the API tweaks for the event source itself, I feel like that simplifies user code nicely. I'll give it another read through tomorrow as well and see if I have more detailed feedback.

src/sys/epoll.rs
```diff
@@ -1,66 +1,126 @@
...
-                    token: unsafe { *token_ptr },
+            .flat_map(|event| {
+                if event.data() == TIMER_DATA {
+                    // We woke up because the high-precision timer fired, we need to disarm it by reading its
+                    // contents to ensure it will be ready for next time
+                    if let Some(ref timer) = self.timer_fd {
+                        // Timer is created in non-blocking mode, and should have already fired anyway, this
+                        // cannot possibly block
+                        let _ = timer.wait();
+                    } else {
```
detly, line 95: Why not do `self.timer_fd.expect("Got an event from etc etc").wait()`?

src/loop_logic.rs
```diff
@@ -220,7 +220,22 @@ impl<'l, Data> EventLoop<'l, Data> {
...
+    }
+
+    /// Create a new event loop in high precision mode
+    ///
+    /// On some platforms it requires to setup more resources to enable high-precision
+    /// (sub millisecond) capabilities, so you should use this constructor if you need
+    /// this kind of precision.
+    ///
+    /// Fails if the initialization of the polling system failed.
+    pub fn try_new_high_precision() -> crate::Result<Self> {
```
detly, line 233: API question: does it make more sense to

- have this be a runtime check ie. return a `Result` or
- be feature gated, so that it doesn't even compile on systems that don't support it

...? In other words, do we expect anyone to do

```rust
let mut event_loop = match EventLoop::try_new_high_precision() {
  Ok(loop) => loop,
  Err(_) => {
    log::warn!("High precision event loop not supported! Falling back on regular old event loop.");
    EventLoop::try_new().expect("Couldn't create event loop")
  }
};
```

Or is it more likely that anyone who _needs_ the HP loop will not be willing to continue if only the regular loop is available?

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
+    {
+        if let Some((my_token, ref wheel)) = self.registration {
+            if my_token != token {
+                return Ok(PostAction::Continue);
+            }
+            let new_deadline = match callback(self.deadline, &mut ()) {
+                TimeoutAction::Drop => return Ok(PostAction::Remove),
+                TimeoutAction::ToInstant(instant) => instant,
+                TimeoutAction::ToDuration(duration) => Instant::now() + duration,
```
detly, line 83: Do you want it to be `Instant::now() + duration` or `self.deadline + duration`? The first would provide more accurate wait-between-action times at the expense of deadlines creeping later and later over a long period; the second would have more jittery callback timing (perhaps) but have a more stable *average* frequency over a long period of time ie. a more precise *number of callbacks* in a given period.

detly: (Some timer APIs I've used actually have a setting for this tradeoff, you might see it as overkill.)

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+            };
+            // If we received an event, we MUST have a valid counter value
+            wheel
+                .lock()
```
detly, line 87: This seems like it could theoretically block the loop. Is there a reason to think it won't?

If so, why not use `try_lock()` and panic if it fails, to effectively `assert!()` that logic?

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+        token_factory: &mut TokenFactory,
+    ) -> crate::Result<()> {
+        self.unregister(poll)?;
+        self.register(poll, token_factory)
+    }
+
+    fn unregister(&mut self, poll: &mut Poll) -> crate::Result<()> {
+        poll.timers.lock().unwrap().cancel(self.counter);
+        self.registration = None;
+        self.counter = 0xFFFFFFFF;
```
detly, line 115: Also a candidate for either (a) `Option<u32>` or (b) `u32::MAX_VALUE` or whatever it is.

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+#[derive(Debug)]
+pub enum TimeoutAction {
+    /// Don't reschedule this timer
+    Drop,
+    /// Reschedule this timer to a given [`Instant`]
+    ToInstant(Instant),
+    /// Reschedule this timer to a given [`Duration`] in the future
+    ToDuration(Duration),
+}
+
```
detly, line 130: I'm still struggling to relate the different parts of this (`TimeoutData`, `TimerWheel`, the `timerfd` in the `Poll`) and how they interact with each other. Would you be able to write a short (perhaps non-doc) comment block that explains it? Even just some bullet points would help (me, and probably future contributors too).

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+            // otherwise this timeout was cancelled, continue looping
+        }
+    }
+
+    pub(crate) fn next_deadline(&self) -> Option<std::time::Instant> {
+        self.heap.peek().map(|data| data.deadline)
+    }
+}
+
+// trait implementations for TimeoutData
```
detly, line 216: Maybe move the debug impl down here as well?

src/sources/timer.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub(crate) fn insert_reuse(&mut self, counter: u32, deadline: Instant, data: T) {
+        self.heap.push(TimeoutData {
+            deadline,
+            data: RefCell::new(Some(data)),
+            counter,
+        });
+    }
+
+    pub(crate) fn cancel(&mut self, counter: u32) {
+        for data in self.heap.iter() {
```
detly, line 181: If you like iterators this compiles to the exact same instructions:

```rust
self.heap
    .iter()
    .find(|data| data.counter == counter)
    .map(|data| data.data.take());
```

src/sys/epoll.rs
```diff
@@ -1,66 +1,126 @@
-use std::os::unix::io::RawFd;
+use std::os::unix::{io::RawFd, prelude::AsRawFd};
```
detly, line 1: Why import this from the prelude instead of doing

```rust
use std::os::unix::io::{RawFd, AsRawFd};
```

?

src/sys/mod.rs
```diff
@@ -186,18 +192,46 @@ impl std::fmt::Debug for Poll {
...
 
     pub(crate) fn poll(
         &mut self,
-        timeout: Option<std::time::Duration>,
+        mut timeout: Option<std::time::Duration>,
     ) -> crate::Result<Vec<PollEvent>> {
-        self.poller.poll(timeout)
+        let now = std::time::Instant::now();
+        // adjust the timeout for the timers
+        if let Some(next_timeout) = self.timers.lock().unwrap().next_deadline() {
```
detly, line 209: I think this is also another potentially blocking operation in the loop - see my suggestion elsewhere to use `try_unlock()` if you have a reason to think this should never block in actual use.

src/sys/mod.rs
```diff
@@ -186,18 +192,46 @@ impl std::fmt::Debug for Poll {
...
+            } else if let Some(deadline) = timeout {
+                timeout = Some(std::cmp::min(deadline, next_timeout - now));
+            } else {
+                timeout = Some(next_timeout - now);
+            }
+        };
+
+        let mut events = self.poller.poll(timeout)?;
+
+        let now = std::time::Instant::now();
```
detly, line 221: There's already a `now` in scope at this point, from L207.

src/loop_logic.rs
i509VCB: If your application is extremely sensitive to timings (namely a compositor trying to present at 360hz, you will hit a fps cap due to timer precision limitations.

Support for timerfd needs to be determined at runtime. Kqueue based polling always work in high precision.

detly: I don't know much about compositors - is hitting that cap something that still allows you to function?

i509VCB: Yeah you could still function, depending on the context it may just be slightly less responsive than you expect or have tearing.

detly: Question answered then.

src/sources/timer.rs
vberger: The current API allows for both, given the current `self.deadline` is passed as argument to the callback, if the user wants `self.deadline + duration`, it can return `ToInstant(deadline + duration)`, if you don't care that much (because the duration is pretty long for example), just return `ToDuration(duration)`. At least that's how I thought about it.

src/sys/mod.rs
vberger: Yes, but given we are after `poller.poll()` returns, a lot of time can have elapsed, the point is to update what `now` is. I'll add a comment making this explicit.

src/sys/epoll.rs
vberger: Well, we shall thank rust-analyzer's auto-fill. :sweat_smile:

src/loop_logic.rs
vberger: > Or is it more likely that anyone who needs the HP loop will not be willing to continue if only the regular loop is available?

I think that, at least you would want to know that the HP loop could not be initialized and be aware that you're going to be running low-precision, possibly to adjust your logic accordingly (setting a flag somewhere, logging a warning, putting a cap on the screen refresh rate you'll allow the user to configure, etc...)

src/sys/mod.rs
vberger: The only other places where this lock is acquired are in the register/reregister/unregister methods of `Timer`, and this lock is acquired there for a short while (just the time to update the timer info). Given the loop and its handles are not threadsafe, I guess this can actually be replaced by a `Rc<RefCell<_>>`. That would make `Timer` non threadsafe as well, but tbh I don't think that is a problem, there is no point in sending a `Timer` between threads before inserting it in an event loop.

src/sources/timer.rs
detly: Ah of course.

src/sys/mod.rs
detly: Oh of course, I missed that.

vberger:
@delty I believe I have addressed all your comments, how does this look now?

detly:
@vberger Beautiful, thanks!

PR #92, opened 2022-04-03
Implement an eventfd-based ping source for Linux.

detly:
This separates and reorganises the underlying mechanics for the ping source, and implements an eventfd-based ping for Linux.

---

~~Currently a draft, but close to final.~~ Partly addresses #15, maybe there's an analogue for BSD-ishes?

Eventfd has a bit of a quirk I didn't think about until I got halfway through this. It gives you *one* fd instead of two like a pipe. This means that if you close that fd, eg. through `CloseOnDrop`, the receiving end's fd becomes invalid and so does the `Generic` source wrapping it. In practice you just get some `EBADFD` errors, but theoretically if the fd is reassigned you could get some extreme weirdness.

I solved this by (a) adding a `dup()` system call at creation to create separate reading and writing fds. Closing a duplicated fd does *not* affect others that are open, solving the first problem.

The other problem is that the source's fd has no way of signalling that the other fd has been closed. Adding another event source to handle this seemed like it would cancel out the performance/complexity benefits of using eventfd in the first place. I have attempted to solve this by using an `AtomicBool` flag wrapped in an `Arc`, keeping the sending handle thread-safe and cloneable.

This sort of begged the question for me though, how does the pipe-based ping handle this? The event-source-code seems to suggest that reading zero bytes from the pipe after it becomes readable means that the other end was closed. But I can't find any documentation saying that closing a pipe end will make the other end become readable in the first place. Anyway, it might be the case that there's no foolproof way to handle this. I'd welcome any suggestions.

Other notes:

- I marked some functions as `#[inline]`, not so much for performance reasons but more to indicate that they're just used as interchangeable pieces in the other helper functions.
- There's a common structure between the pipe and eventfd files (`send_ping()`, `drain_ping()` etc) even though these are private functions. It's not a suggestion that this structure be enforced. It's just that breaking those functions out of the pipe impl helped me build them back up for the eventfd impl. If you prefer them back in the event source, that's fine by me.
- I will add some of these notes in as comments, but perhaps after feedback in case there are any fundamental changes.
- This PR adds [cfg-if](https://crates.io/crates/cfg-if) as a dependency. It should have no runtime cost. I think it helps the implementation switching a bit.

Docs/links:
- [eventfd](https://man7.org/linux/man-pages/man2/eventfd.2.html)
- [pipe2](https://man7.org/linux/man-pages/man2/pipe.2.html)
- SO: [Do I get a notification from epoll when a fd is closed?](https://stackoverflow.com/questions/43034535/do-i-get-a-notification-from-epoll-when-a-fd-is-closed)

### DIFF
--- a/src/sources/ping/eventfd.rs
+++ b/src/sources/ping/eventfd.rs
@@ -1,0 +1,189 @@
+//! Eventfd based implementation of the ping event source.
+//!
+//! # Implementation notes
+//!
+//! The eventfd is a much lighter signalling mechanism provided by the Linux
+//! kernel. Rather than write an arbitrary sequence of bytes, it only has a
+//! 64-bit counter.
+//!
+//! To avoid closing the eventfd early, we wrap it in a RAII-style closer
+//! `CloseOnDrop` in `make_ping()`. When all the senders are dropped, another
+//! wrapper `FlagOnDrop` handles signalling this to the event source, which is
+//! the sole owner of the eventfd itself. The senders have weak references to
+//! the eventfd, and if the source is dropped before the senders, they will
+//! simply not do anything (except log a message).
+//!
+//! To differentiate between regular ping events and close ping events, we add 2
+//! to the counter for regular events and 1 for close events. In the source we
+//! can then check the LSB and if it's set, we know it was a close event. This
+//! only works if a close event never fires more than once.
+
+use std::{os::unix::io::RawFd, sync::Arc};
+
+use nix::sys::eventfd::{eventfd, EfdFlags};
+use nix::unistd::{read, write};
+
+use super::{CloseOnDrop, PingError};
+use crate::{
+    generic::Generic, EventSource, Interest, Mode, Poll, PostAction, Readiness, Token, TokenFactory,
+};
+
+// These are not bitfields! They are increments to add to the eventfd counter.
+// Since the fd can only be closed once, we can effectively use the
+// INCREMENT_CLOSE value as a bitmask when checking.
+const INCREMENT_PING: u64 = 0x2;
+const INCREMENT_CLOSE: u64 = 0x1;
+
+#[inline]
+pub fn make_ping() -> std::io::Result<(Ping, PingSource)> {
+    let read = eventfd(0, EfdFlags::EFD_CLOEXEC | EfdFlags::EFD_NONBLOCK)?;
+
+    // We only have one fd for the eventfd. If the sending end closes it when
+    // all copies are dropped, the receiving end will be closed as well. We need
+    // to make sure the fd is not closed until all holders of it have dropped
+    // it.
+
+    let fd_arc = Arc::new(CloseOnDrop(read));
+
+    let ping = Ping {
+        event: Arc::new(FlagOnDrop(Arc::clone(&fd_arc))),
+    };
+
+    let source = PingSource {
+        event: Generic::new(read, Interest::READ, Mode::Level),
+        _fd: fd_arc,
+    };
+
+    Ok((ping, source))
+}
+
+// Helper functions for the event source IO.
+
+#[inline]
+fn send_ping(fd: RawFd, count: u64) -> std::io::Result<()> {
+    assert!(count > 0);
+    match write(fd, &count.to_ne_bytes()) {
+        // The write succeeded, the ping will wake up the loop.
+        Ok(_) => Ok(()),
+
+        // The counter hit its cap, which means previous calls to write() will
+        // wake up the loop.
+        Err(nix::errno::Errno::EAGAIN) => Ok(()),
+
+        // Anything else is a real error.
+        Err(e) => Err(e.into()),
+    }
+}
+
+#[inline]
+fn drain_ping(fd: RawFd) -> std::io::Result<u64> {
+    // The eventfd counter is effectively a u64.
+    const NBYTES: usize = 8;
+    let mut buf = [0u8; NBYTES];
+
+    match read(fd, &mut buf) {
+        // Reading from an eventfd should only ever produce 8 bytes. No looping
+        // is required.
+        Ok(NBYTES) => Ok(u64::from_ne_bytes(buf)),
+
+        Ok(_) => unreachable!(),
+
+        // Any other error can be propagated.
+        Err(e) => Err(e.into()),
+    }
+}
+
+// The event source is simply a generic source with one of the eventfds.
+#[derive(Debug)]
+pub struct PingSource {
+    event: Generic<RawFd>,
+
+    // This is held only to ensure that there is an owner of the fd that lives
+    // as long as the Generic source, so that the fd is not closed unexpectedly
+    // when all the senders are dropped.
+    _fd: Arc<CloseOnDrop>,
+}
+
+impl EventSource for PingSource {
+    type Event = ();
+    type Metadata = ();
+    type Ret = ();
+    type Error = PingError;
+
+    fn process_events<C>(
+        &mut self,
+        readiness: Readiness,
+        token: Token,
+        mut callback: C,
+    ) -> Result<PostAction, Self::Error>
+    where
+        C: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
+    {
+        self.event
+            .process_events(readiness, token, |_, &mut fd| {
+                let counter = drain_ping(fd)?;
+
+                // If the LSB is set, it means we were closed. If anything else
+                // is also set, it means we were pinged. The two are not
+                // mutually exclusive.
+                let close = (counter & INCREMENT_CLOSE) != 0;
+                let ping = (counter & (u64::MAX - 1)) != 0;
+
+                if ping {
+                    callback((), &mut ());
+                }
+
+                if close {
+                    Ok(PostAction::Remove)
+                } else {
+                    Ok(PostAction::Continue)
+                }
+            })
+            .map_err(|e| PingError(e.into()))
+    }
+
+    fn register(&mut self, poll: &mut Poll, token_factory: &mut TokenFactory) -> crate::Result<()> {
+        self.event.register(poll, token_factory)
+    }
+
+    fn reregister(
+        &mut self,
+        poll: &mut Poll,
+        token_factory: &mut TokenFactory,
+    ) -> crate::Result<()> {
+        self.event.reregister(poll, token_factory)
+    }
+
+    fn unregister(&mut self, poll: &mut Poll) -> crate::Result<()> {
+        self.event.unregister(poll)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Ping {
+    // This is an Arc because it's potentially shared with clones. The last one
+    // dropped needs to signal to the event source via the eventfd.
+    event: Arc<FlagOnDrop>,
+}
+
+impl Ping {
+    /// Send a ping to the `PingSource`.
+    pub fn ping(&self) {
+        if let Err(e) = send_ping(self.event.0 .0, INCREMENT_PING) {
+            log::warn!("[calloop] Failed to write a ping: {:?}", e);
+        }
+    }
+}
+
+/// This manages signalling to the PingSource when it's dropped. There should
+/// only ever be one of these per PingSource.
+#[derive(Debug)]
+struct FlagOnDrop(Arc<CloseOnDrop>);
+
+impl Drop for FlagOnDrop {
+    fn drop(&mut self) {
+        if let Err(e) = send_ping(self.0 .0, INCREMENT_CLOSE) {
+            log::warn!("[calloop] Failed to send close ping: {:?}", e);
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   associated event callback.
 - **Breaking:** The minimum supported Rust version is now 1.53.0
 - Introduce `EventLoop::try_new_high_precision()` for sub-millisecond accuracy in the event loop
+- The `PingSource` event source now uses an `eventfd` instead of a pipe on Linux.
 
 ## 0.9.2 -- 2021-12-27
 
--- a/src/sources/ping.rs
+++ b/src/sources/ping.rs
@@ -4,20 +4,31 @@
 //! [`Ping::ping`](Ping#method.ping) method is called. If the event source is pinged multiple times
 //! between a single dispatching, it'll only generate one event.
 //!
-//! This event loop is a simple way of waking up the event loop from an other part of your program
+//! This event source is a simple way of waking up the event loop from an other part of your program
 //! (and is what backs the [`LoopSignal`](crate::LoopSignal)). It can also be used as a building
 //! block to construct event sources whose source of event is not file descriptor, but rather an
 //! userspace source (like an other thread).
 
-use std::{os::unix::io::RawFd, sync::Arc};
+use nix::unistd::close;
+use std::os::unix::io::RawFd;
 
-use nix::{
-    fcntl::OFlag,
-    unistd::{close, read, write},
-};
+// The ping source has platform-dependent implementations provided by modules
+// under this one. These modules should expose:
+// - a make_ping() function
+// - a Ping type
+// - a PingSource type
+//
+// See eg. the pipe implementation for these items' specific requirements.
 
-use super::generic::Generic;
-use crate::{EventSource, Interest, Mode, Poll, PostAction, Readiness, Token, TokenFactory};
+#[cfg(target_os = "linux")]
+mod eventfd;
+#[cfg(target_os = "linux")]
+use eventfd as platform;
+
+#[cfg(not(target_os = "linux"))]
+mod pipe;
+#[cfg(not(target_os = "linux"))]
+use pipe as platform;
 
 /// Create a new ping event source
 ///
@@ -25,43 +36,7 @@ use crate::{EventSource, Interest, Mode, Poll, PostAction, Readiness, Token, Tok
 /// event loop, and a [`PingSource`], which you can insert in your event loop to
 /// receive the pings.
 pub fn make_ping() -> std::io::Result<(Ping, PingSource)> {
-    #[cfg(not(target_os = "macos"))]
-    let (read, write) = {
-        use nix::unistd::pipe2;
-
-        pipe2(OFlag::O_CLOEXEC | OFlag::O_NONBLOCK)?
-    };
-
-    // macOS does not have pipe2, but we can emulate the behavior of pipe2 by setting the flags after calling pipe.
-    #[cfg(target_os = "macos")]
-    let (read, write) = {
-        use nix::{
-            fcntl::{fcntl, FcntlArg},
-            unistd::pipe,
-        };
-
-        let (read, write) = pipe()?;
-
-        let read_flags = OFlag::from_bits_truncate(fcntl(read, FcntlArg::F_GETFD)?)
-            | OFlag::O_CLOEXEC
-            | OFlag::O_NONBLOCK;
-        let write_flags = OFlag::from_bits_truncate(fcntl(write, FcntlArg::F_GETFD)?)
-            | OFlag::O_CLOEXEC
-            | OFlag::O_NONBLOCK;
-
-        fcntl(read, FcntlArg::F_SETFL(read_flags))?;
-        fcntl(write, FcntlArg::F_SETFL(write_flags))?;
-
-        (read, write)
-    };
-
-    let source = PingSource {
-        pipe: Generic::new(read, Interest::READ, Mode::Level),
-    };
-    let ping = Ping {
-        pipe: Arc::new(CloseOnDrop(write)),
-    };
-    Ok((ping, source))
+    platform::make_ping()
 }
 
 /// The ping event source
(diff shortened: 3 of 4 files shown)


kchibisov:
>The event-source-code seems to suggest that reading zero bytes from the pipe after it becomes readable means that the other end was closed.

From `read(2)` it says "On  success,  the  number  of bytes read is returned (zero indicates end of file), and the file position is advanced by this number."

Also, the way it works is due to closing triggering wakeup due to fd event leading to read from pipe read fd, while the write end is closed resulting in getting `0`, since we've reached EOF.

detly:
Oh thanks, now I know! Unfortunately it doesn't apply to eventfd AFAICT. I don't think closing the `dup`-ed triggers an event either. I tried to test this behaviour (and its absence), but couldn't come up with a test for it.

detly:
> Also, the way it works is due to closing triggering wakeup due to fd event leading to read from pipe read fd, while the write end is closed resulting in getting 0, since we've reached EOF.

I think there's still an issue with this though. The read loop in the event source will drain all bytes from the pipe on an event, probably as it should. But if you imagine a sequence of events on the sender: ping, followed by drop/close, *before* the next dispatch cycle... I think  you'll get one event, read the bytes, finish on the `Ok(0)` arm, `break`, and keep the ping source in the loop. Right?

kchibisov:
>I think you'll get one event, read the bytes, finish on the Ok(0) arm, break, and keep the ping source in the loop. Right?

With your new code yes, the old code handled that fine. So you should follow its behavior.

detly:
I didn't think my new code changed the functionality of the pipe-based ping source?

kchibisov:
>I didn't think my new code changed the functionality of the pipe-based ping source?

I mean you clearly did, since you're not dropping event source in the case you've described above, while the current ping source will drop, look at the loop it had before with `match read`.

detly:
Okay I wrote this out and see I dropped the `PostAction::Remove` from [L102](https://github.com/Smithay/calloop/blob/7835e53b61fae03487da050600b20af3d749b5a3/src/sources/ping.rs#L102). Let me fix that.

detly:
The pipe-based ping behaviour is restored. The eventfd source still has that issue though. I think adding a final ping send to the drop handler for the sending handle might work.

kchibisov:
>The other problem is that the source's fd has no way of signalling that the other fd has been closed. Adding another event source to handle this seemed like it would cancel out the performance/complexity benefits of using eventfd in the first place. I have attempted to solve this by using an AtomicBool flag wrapped in an Arc, keeping the sending handle thread-safe and cloneable.

I don't see major issue with it? The point is to have less file descriptors open, no? if you share bool flag between sender and receiver it should be fine? Also, you could bench pipe and eventfd I guess, not sure what the way of doing so should be.

detly:
> I don't see major issue with it?

I'm not sure what detail you mean here. The problem I'm trying to avoid is event sources that don't return `PostAction::Remove` when all the sending handles have been dropped. The boolean flag itself is not enough, because it will not necessarily wake up the event source to read it and see that it needs removal.

> Also, you could bench pipe and eventfd I guess, not sure what the way of doing so should be.

I don't really care about performance, in the vague sense of the word, but maybe Calloop and its other users have a specific idea of things they want to avoid. `Ping` is such a fundamental building block of the event loop that I'd really like to avoid unnecessary increases in resource usage (especially since eventfd is meant to do the opposite), whether that's processing, memory, or fds.

kchibisov:
>The boolean flag itself is not enough, because it will not necessarily wake up the event source to read it and see that it needs removal.

You can ping in drop right before setting bool flag, check that bool flag is set and instead of calling callback drop the source.

kchibisov:
And in case you ping and drop right away you can pass bitmask along the line e.g. `READ | DROP` instead of plain boolean flag.

detly:
> And in case you ping and drop right away you can pass bitmask along the line e.g. READ | DROP instead of plain boolean flag.

I was implementing this and I realised we can do this purely with the event counter! Write 2 for "real" ping events, and 1 for drop events. Then we just have to bit-test the LSB and if it's set, we're closed.

detly:
See latest for this implementation and remarks. It's still a draft because I'd really like to add a test for this. I'll have time for that in a few hours.

detly:
I've added two tests for (a) ping source removal and also (b) simultaneous ping and source removal. They fail for my previous code and pass for the fixed code.

kchibisov:
>I was implementing this and I realised we can do this purely with the event counter! Write 2 for "real" ping events, and 1 for drop events. Then we just have to bit-test the LSB and if it's set, we're closed.

What if you concurrently drop 2 ping sources? You'll add `2` from 2 drop impls and it won't be dropped? I was also thinking about using counter don't get me wrong it, but I wasn't able to figure out how you'd do that in a nice way... You could probably use `Weak` for wake up sources and check for weak ref count, so if it's zero everything is dropped.

src/sources/ping/eventfd.rs
```diff
@@ -0,0 +1,190 @@
...
+const INCREMENT_PING: u64 = 0x2;
+const INCREMENT_CLOSE: u64 = 0x1;
+
+#[inline]
+pub fn make_ping() -> std::io::Result<(Ping, PingSource)> {
+    let read = eventfd(0, EfdFlags::EFD_CLOEXEC | EfdFlags::EFD_NONBLOCK)?;
+    // We only have one fd for the eventfd. If the sending end closes it when
+    // all copies are dropped, the receiving end will be closed as well. We can
+    // avoid this by duplicating the FD.
+    let write = dup(read)?;
```
vberger, line 40: If this fails we are leaking a file descriptor

src/sources/ping/eventfd.rs
```diff
@@ -0,0 +1,190 @@
...
+        // Anything else is a real error.
+        Err(e) => Err(e.into()),
+    }
+}
+
+#[inline]
+fn drain_ping(fd: RawFd) -> std::io::Result<u64> {
+    // The eventfd counter is effectively a u64.
+    const NBYTES: usize = std::mem::size_of::<u64>();
+    let mut buf = [0u8; NBYTES];
```
vberger, line 75: Given this is documented to be a `u64` (and not an `usize`), the size is not platform-dependent, so that seems like a needlessly complex way to write `8` tbh (even though I guess the compile-time cost is negligible...)

src/sources/ping/eventfd.rs
```diff
@@ -0,0 +1,190 @@
...
+                    callback((), &mut ());
+                }
+
+                let action = if close {
+                    PostAction::Remove
+                } else {
+                    PostAction::Continue
+                };
+
+                Ok(action)
```
vberger, line 130:
```suggestion
                if close {
                   Ok(PostAction::Remove)
                } else {
                    Ok(PostAction::Continue)
                }
```

vberger (commented):
That looks very nice, thanks a lot :+1: I just have a few nitpicks.

detly:
> What if you concurrently drop 2 ping sources? You'll add 2 from 2 drop impls and it won't be dropped?

Eventfd works as an atomic counter, not a pipe, so the numbers are added. The result of that sequence would be that the event source reads 5, and the masking does the rest.

kchibisov:
>Eventfd works as an atomic counter, not a pipe, so the numbers are added. The result of that sequence would be that the event source reads 5, and the masking does the rest.

That's what I'm saying. `Ping + Ping + Drop + Drop` is `2 + 2 + 1 + 1 == 6` is the same as `Ping + Ping + Ping`.

src/sources/ping/eventfd.rs
kchibisov: Can we use `Arc` instead of `dup` here? I don't see a reason to dup it. We could have `Weak<Fd>` pointers in senders and check in receiver ref count and drop fd it goes to zero.

src/sources/ping/eventfd.rs
detly: Zero, in fact because it's a `const fn` :wink:

Look, I am known to be pathological about this, but if I can find a way to avoid telling a computer a number I will. I will happily change it to 8 though because I recognise that I have a problem. (Do you object to the constant `NBYTES` itself or just the `size_of` call?)

kchibisov:
On the other side, I don't really think that automatic drop should be performed in the first place? The close should occur in the `Drop` impl. If the source is being removed from the event loop, the `Drop` will be called meaning that we can safely close it, no? There shouldn't be any ref counting in the first place.

If we've dropped all the senders we shouldn't remove source, since I think it could make sense to have ability to recreate ping wake up sources later on?

Anyway, it's more up to @vberger I guess wrt ping design, I'm mostly saying that automatic drop is a bit intense here.

detly:
> The Drop is twice, due to parallel drop of 2 sources. Should work with any even amount of sources dropped at the same time.

Note that the `FlagOnDrop` is wrapped in an `Arc`, so even across threads the drop can only happen once. If it didn't, there would actually be two bugs: (1) is exactly what you describe, and (2) is a double close on the wrapped fd. (2) would also apply to the pipe-based source I think due to the similar `CloseOnDrop` wrapper.

src/sources/ping/eventfd.rs
vberger: > Zero, in fact because it's a const fn

Hence why I was talking about *compile-time* cost. ^^

I don't have any strong feelings about this tbh, it just kinda surprised me to see this. No objection to keeping it like this if you strongly prefer it.

vberger:
> If we've dropped all the senders we shouldn't remove source, since I think it could make sense to have ability to recreate ping wake up sources later on?

Isn't that covered by just creating a new `PingSource` and inserting it in the event loop? I have a feeling that if you want to keep a specific ping source (and its callback) alive, you just keep at least one copy of its associated `Ping` around (because you plan to use it at some point anyway)... :thinking:

src/sources/ping/eventfd.rs
detly: Oh oops. Anyway, I am not married to it, I'll change it when I do @kchibisov's suggestion for the duplicated fd.

detly:
> If we've dropped all the senders we shouldn't remove source, since I think it could make sense to have ability to recreate ping wake up sources later on?

The removal via `process_events()` return value helps a lot with dynamically created sources and senders. I don't know if you've seen #45 but that explains the motivation for it, and it's been useful for eg. channels.

> The close should occur in the Drop impl. If the source is being removed from the event loop, the Drop will be called meaning that we can safely close it, no?

Note that this is still the case: if `process_events()` returns `Remove`, the loop logic will remove it internally, drop it, and then the drop handler does the close.

src/sources/ping/eventfd.rs
detly: > We could have Weak<Fd> pointers in senders and check in receiver ref count and drop fd it goes to zero.

This would work *if* we can be sure the weak count (ie. number of senders) will never go to zero *and then* come back up again. Can anyone think of a situation where that might happen? (I can't.)

vberger: I'm not quite sure I get how your suggestion would work out @kchibisov

Do you mean checking the refcount every time the source is woken up, and drop the source if it reaches zero? It seems to me that it implies waking up the source every time a `Ping` is dropped, and how would we be able to distinguish that from a regular `ping()`? Dropping a `Ping` handle would cause spurious events on the event source, no?

detly: Yeah I just tried this approach and hit this exact problem. Spurious events aren't such a big deal for me (you get a lot of them with the async executor for example, depending on what's in there), but yes, there's that. And then there's the "what kind of wakeup was this" issue, which means we'd still need the `FlagOnDrop` trick, which means two levels of `Arc` (on for sharing the fd between senders and source, and one for sharing it between just the senders).

kchibisov: @vberger I've also realized that it's a **problem** later on. In general I don't want to have lots of fd's. The issue is that identifying that certain ping should dispatch callback, but some other don't is complicated, and I'm not sure that we can solve it without wakeups.

I think what we can do is to pass some value for every ping, but when the drop is performed the value should be zero. So if we woken up and the value is zero then we shouldn't dispatch callback, and if ref count went to zero it means that drop should be performed. And every non zero is drained on read, so it should be fine.

kchibisov: @detly have you tried writing `1` for regular pings and `0` for drop ones? So you could check if the value is `zero` in wakeup => no callback call, and it ref count went to zero as well it means that drop should be done?

I think it could work without extra flags, just not sure if passing zero is allowed.

detly: @kchibisov Alas passing 0 is not allowed, or rather it is, but it won't trigger an event on the `eventfd`.

detly: Oh, there *is* a semaphore mode for the eventfd. Let me take a look at that.

detly: Nope, it's more about semaphore *events* not ownership, I don't think it helps us.

detly: At any rate, my recent push fixes the fd leak, but I'm happy to rework the sender management if there's a better way.

kchibisov: If you're insisting on using `dup` and not pass it around by arc, be aware that `dup` discards flags, like `EFD_CLOEXEC`, so you should set them manually. Not sure about other apis nix provide.

detly: Good catch, I will fix that too.

I'm not so much insisting on it, more that I can't really get a recount approach to work without introducing some other problem (loop wakeups, race conditions, or API ugliness are the main ones).

kchibisov: I mean you don't need ref count to work, you can use your approach, but use arc instead of dup. The only change you'll add is remove close from client, and check refcount before dropping.

detly: > The only change you'll add is remove close from client, and check refcount before dropping.

What happens if the source is dropped before the senders?

kchibisov: The `Weak` will tell you that?

detly: > The Weak will tell you that? It's like weak for a reason...

Right, so now we're back to - how do you unambiguously signal to the source that a sender has been dropped? Dropping the last `Weak` ref won't trigger a drop, because the source needs to hold on to the original `Arc`.

detly: Maybe wrap the `Weak` in another `Arc<FlagOnDrop>` that's shared amongst all the senders? That could work...

detly: Ah, yes, I think I have something now.

detly: Okay. Okay. I think this last push might satisfy everyone. Thanks for your persistence @kchibisov, this really should be a lot safer. I added another test to check the case of multiple senders and dropping them in between dispatches.

vberger: Do I understand that the only point of using `Weak` here is so that the FD is closed as soon as the `PingSource` is dropped, rather than once the `PingSource` and all `Ping`s are dropped?

detly: Sort of. It was partly that, and also so that simply holding on to a `Ping` did not keep the `PingSource`'s underlying eventfd alive.

I see your point though, the implementation would be simpler if it was `Arc`s all the way.

detly: Latest version has full `Arc`s all the way through. I've left the intermediate commits in in case you want to compare, but I'll squash them when we're happy with it.

vberger: I don't think that's a blocking point though, that was just to make sure I understand the point.

detly: I think I prefer the full-Arc version, it *is* simpler in general.

vberger (approved):
LGTM, maybe could you just add a note about this change in the changelog?

Cargo.toml
```diff
@@ -15,6 +15,7 @@ readme = "README.md"
 codecov = { repository = "Smithay/calloop" }
 
 [dependencies]
+cfg-if = "1.0.0"
```
kchibisov, line 18: Do we really need cfg if? Right now we're handling cases without it just fine. Not sure it worth?

src/sources/ping/eventfd.rs
```diff
@@ -0,0 +1,189 @@
...
+    fn unregister(&mut self, poll: &mut Poll) -> crate::Result<()> {
+        self.event.unregister(poll)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct Ping {
+    // This is an Arc because it's potentially shared with clones. The last one
+    // dropped needs to signal to the event source via the eventfd.
+    event: Arc<FlagOnDrop>,
```
kchibisov, line 166: There is no need for nested arcs? The FlagOnDrop is already an arc? Otherwise whats the purpose of nesting arcs?

Cargo.toml
detly: Sure, it was necessary when I split out the macos pipe vs. non-macos pipe2 implementations, but probably not any more.

src/sources/ping/eventfd.rs
detly: If you clone the `FlagOnDrop` throughout all the senders, a different one will be dropped every time a sender is dropped. This will (a) wake the source up every time with `send_ping()` in its drop handler and (b) hit the problem you mentioned elsewhere, where the counter may be incremented by 2 (or an even number) instead of 1 if two are dropped in between loop dispatches.

There must *strictly* be one `FlagOnDrop` in existence for any given `PingSource` for the signalling to work.

To put it another way: `Arc`s are for shared ownership.
- The inner level of `Arc` is for shared ownership of the eventfd by both the `PingSource` and all `Ping`s, original or cloned.
- The outer level of `Arc` is for shared ownership of the *single* "thing that adds 1 to the eventfd counter" amongst all the `Ping` clones.

They do not conceptually share ownership of the same thing, so you need two levels.

detly: Here's a quick picture. Note how the top `Arc` (the one containing `FlagOnDrop`) is not shared with `PingSource`. This makes `FlagOnDrop` uniquely responsible for adding numbers to the eventfd counter.

arcs

detly:
> LGTM, maybe could you just add a note about this change in the changelog?

Which change? Github won't show me what this refers to.

kchibisov:
Actually, when dropping ping, can you check the ref count. And if it is the last ping to drop you write 1, otherwise you just silently drop? I think it should work and simplify a lot.

kchibisov:
Thats some pseudo code but you should get an idea

```rust
struct Ping {
    fd: Weak<EventFd>,
}
impl Ping {
    fn ping() {
        self.fd.write(2);
    }
}

impl Drop for Ping {
    fn drop(&mut self) {
        if fd.weak_count() == 1 {
            self.fd.write(1);
        }
    }
}

impl PingSource {
    fn process_events() {
        // ....
        bool should_close = false;
        match fd.read() {
            Ok(n) if n == 1 {
                should_close = true;
            }
            Ok(n) {
                callback();
                should_close = n % 2 == 0;
            }
            _ => // nothing...
        }
        if should_close {
            std::mem::drop(self.fd);
        }
        // ....
    }
}
```

vberger:
```rust
impl Drop for Ping {
    fn drop(&mut self) {
        if fd.weak_count() == 1 {
            self.fd.write(1);
        }
    }
}
```

I think that part is racy, if the last two `Ping` are dropped concurrently drop two threads, both could see `weak_count == 2` and decide to not write.

kchibisov:
>I think that part is racy, if the last two Ping are dropped concurrently drop two threads, both could see weak_count == 2 and decide to not write.

Yeah, and I don't see a way to decrement and get value of ref count. There could be a `Mutex` just for drop to ensure thread safety.

Since right now if you drop from 2 threads at the same time you won't drop, since you write `0x1` from both resulting in `0x2`, since it's **counter**.

vberger:
> Since right now if you drop from 2 threads at the same time you won't drop, since you write 0x1 from both resulting in 0x2, since it's counter.

I'm not sure what you mean by "right now", the PR as it is current does not have this issue, thanks to the two layers of `Arc`

kchibisov:
Yeah, maybe double arc isn't that bad.

vberger:
In any case, I think this is deep into optimization land, and worst case this can always be improved in a future PR if we find a nice way to do it.

detly:
> Since right now if you drop from 2 threads at the same time you won't drop, since you write 0x1 from both resulting in 0x2, since it's counter.

FYI this is the promise of both an `Arc` vs `Rc`, and of `Drop` - in `Arc`s case, the count is changed atomically even from multiple threads, and for `Drop`, the drop handler is only ever called once (otherwise there would be a lot of unsafe code out there). So the write of `1` can only ever happen once, as long as there is a unique `FlagOnDrop` for every `PingSource`.

Re. raciness, see this note from the [`Arc` docs](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.weak_count):

> This method by itself is safe, but using it correctly requires extra care. Another thread can change the weak count at any time, including potentially between calling this method and acting on the result.

detly:
@vberger Do you want me to also add a note about MSRV change to 1.53 in the changelog, or would you rather keep it in a separate PR?

detly:
Changelog updated.

vberger:
Alright thanks, let's merge this. :+1:

PR #93, opened 2022-04-15
Fill in book chapters on generic and ping sources.

detly:
I filled in a couple of the stub chapters (generic and ping). There's more to do, but I figured I'd push what I have now.

As an aside, the timer chapter is a bit of an odd one out, because it's very detailed compared to what I've written for these two. I'm not sure what would be better - having detailed examples for all of calloop's sources (which would be a lot of work), or maybe moving the timer example to the end (like maybe before the ZeroMQ example) and just having a summary there.


vberger:
Thanks!

A large part of the contents of the timer chapter is completely outdated following the timer changes, so it'll need to be mostly rewritten anyway.

doc/src/ch02-01-generic.md
```diff
@@ -1 +1,23 @@
...
+
+For example on Linux, fd-based interfaces are available for GPIO, I2C, USB, UART, network interfaces, timers and many other systems. Integrating these into calloop starts with obtaining the appropriate fd, creating a `Generic` event source from it, and building up a more useful, abstracted event source around that. A detailed example of this is [given for ZeroMQ](ch04-00-a-full-example-zeromq.md). 
+
+## Creating a Generic source
+
+Creating a `Generic` event source requires three things:
+- The fd itself, of course
+- The ["interest"](api/calloop/struct.Interest.html) you have in this descriptor — this means, whether you want to generate events when the fd becomes readable, writeable, or both
+- The ["mode"](api/calloop/enum.Mode.html) of event generation - level triggered or edge triggered. If you are familiar with hard descriptor languages, you can think of this as the sensitivity.
```
vberger, line 11: > If you are familiar with hard descriptor languages, you can think of this as the sensitivity.

I am not and that means nothing to me. Could we maybe instead make the explanation more explicit?

doc/src/ch02-01-generic.md
```diff
@@ -1 +1,23 @@
...
+
+## Ownership
+
+It's important to remember that file descriptors have no concept of ownership attached to them in Rust — they are simply bare integers. Typically (eg. in the standard library) they would be an underlying implementation detail of another type that *did* encode ownership somehow, but in order to integrate so closely with the systems they expose, we must use them (and their platform APIs) directly.
+
+This means you need to do at least two things:
+- follow the rules of the API you obtain the fd from
+- wrap them in a type that manages ownership appropriately
+
+For example, the `Ping` event source in calloop uses a reference-counted `CloseOnDrop` wrapper to ensure that the underlying fd is closed when all clones of the ping's generic event source are, themselves, dropped.
```
vberger, line 23: I'm not a huge fan of talking about `calloop`'s internals in the documentation like this. Let's maybe instead encourage using appropriate types, like for example illustrate how `Generic<Socket>` will close the socket on drop?

And generally, `Generic` is not only about fds, but about types which are backed by fds.

doc/src/ch02-01-generic.md
detly: Well, I meant to write "hardWARE descriptor langauges such as VHDL", but it's not really necessary. In retrospect I don't think it adds much.

doc/src/ch02-01-generic.md
detly: > like for example illustrate how Generic<Socket> will close the socket on drop?

Ah that's a good point. I will redo this part.

detly:
Quick question — I have the following sentence:

> For example on Linux, fd-based interfaces are available for GPIO, I2C, USB, UART, network interfaces, timers and many other systems.

If you're familiar with BSD systems, can you tell me if that's true for them as well?

doc/src/ch02-01-generic.md
detly: I redid this part with more emphasis on using wrapper types, with an example from the standard library. I didn't mention the `OwnedFd` and `BorrowedFd` types because they're still unstable, but when they stabilise they might warrant a mention here too.

vberger:
> If you're familiar with BSD systems, can you tell me if that's true for them as well?

I'm not really familiar with BSD systems, sorry.

doc/src/ch02-01-generic.md
```diff
@@ -1 +1,27 @@
...
+Creating a `Generic` event source requires three things:
+- The fd itself, of course, possibly in a wrapper type that implements `AsRawFd`
+- The ["interest"](api/calloop/struct.Interest.html) you have in this descriptor — this means, whether you want to generate events when the fd becomes readable, writeable, or both
+- The ["mode"](api/calloop/enum.Mode.html) of event generation - level triggered or edge triggered
+
+The easiest constructor to use is the [`new()`](api/calloop/generic/struct.Generic.html#method.new) method, but if you need control over the [associated error type](ch02-06-errors.md) there is also [`new_with_error()`](api/calloop/generic/struct.Generic.html#method.new_with_error).
+
+## Ownership and AsRawFd wrappers
+
+It's important to remember that file descriptors by themselves have no concept of ownership attached to them in Rust — they are simply bare integers. Dropping them does close the resource they refer to, and copying them does not carry information about how many there are.
```
vberger, line 19:
```suggestion
It's important to remember that file descriptors by themselves have no concept of ownership attached to them in Rust — they are simply bare integers. Dropping them does not close the resource they refer to, and copying them does not carry information about how many there are.
```

detly:
Gah, sorry for leaving this, we all got COVID so I've been out of action for a while. Did you fix these elsewhere or do you need me to make a PR with the amendments?

Good work on the release btw.

vberger:
I fixed that in an other commit, no problem. :+1: Hope you're getting better.

PR #102, opened 2022-06-20
channel: Add (try_)recv proxy methods

Drakulix:


vberger (approved):

vberger:
Could you add a changelog entry as well?

Issue #104, opened 2022-06-29
API: TransientSource requires the loop to run again before the source can be replaced


detly:
Even though I wrote the [`TransientSource` wrapper](https://github.com/Smithay/calloop/blob/613c839be75ba4f172d089bff0a94e5d3dfa3232/src/sources/transient.rs), I've hit a bit of a API/design problem that I could use some help with.

I'll use a Timer as an example, but this is not specifically a problem with timers, I have the same problem with custom sources that fire a finite number of times, expire and need to be replaced.

Let's say I have a source with a timer, but I don't want to start the timer unless something else happens. In the struct I might have:

```rust
struct TimedThing {
    is_sending: bool,
    // Only runs if is_sending is true
    timer: calloop::timer::Timer,
}
```

My constructor then has this problem:

```rust
impl TimedThing {
    fn new() -> Self {
        Self {
            is_sending: false,
            timer: todo!("?????"),
        }
    }
}
```

This is more or less what `TransientSource` is designed for, so I try:

```rust
struct TimedThing {
    is_sending: bool,
    // Only runs if is_sending is true
    timer: calloop::transient::TransientSource<calloop::timer::Timer>,
}
```

```rust
impl TimedThing {
    fn new() -> Self {
        Self {
            is_sending: false,
            timer: calloop::transient::TransientSource::None,
        }
    }
}
```

...and then at some point call `self.timer = Timer::from_duration(...);` when `is_sending` becomes true. **Except that** this only works if the timer does not already exist. If it *does* exist, this will leak ~~an epoll entry~~ whatever underlying thing keeps track of registration ([wheel entry](https://github.com/Smithay/calloop/blob/613c839be75ba4f172d089bff0a94e5d3dfa3232/src/sources/timer.rs#L136) for Timers, epoll entry for fd-based sources), because the previous source will be dropped without being unregistered. 

That's the problem with the `TransientSource` design that I'm trying to solve ie. there's no way to unregister a source without waiting for another iteration of the event loop.

Here's an attempt to work around it (pseudo-ish code):

```rust
fn process_events(...) {
    if (turn_off_sending) {
        self.is_sending = false;
        self.timer = TransientSource::None;
        // Oh no! Even if we return PostAction::Reregister, self.timer
        // no longer owns the timer we just replaced! It has no way to
        // unregister it now. If this were an event source that generated
        // continuous events
    }
}
```

[...]

PR #106, opened 2022-07-04
Mitigate transient source leak

detly:
This addresses #104. Yes, it doubles the size of the type, but I'm inclined to say that's not a huge deal, since you'd need more than that space to do this correctly at a higher level.

The two extra tests are a bit dense, but they basically test the edge cases that led to this in the first place. I have *not* squashed every commit - I've broken it into three parts:

- a minor API change
- the "before" state - a similar test, it passes, but shows the complexity required to do things correctly
- the "after" state - what you see here

Changelog update to follow.

### DIFF
--- a/src/sources/transient.rs
+++ b/src/sources/transient.rs
@@ -97,15 +97,13 @@
 /// inner source, even if it actually needs to be unregistered or initially
 /// registered.
 ///
-/// ## Replacing or removing `TransientSource`s without leaking
+/// ## Replacing or removing `TransientSource`s
 ///
-/// It is possible to leak registration if you bypass the API of a
-/// `TransientSource`. "Leak registration" means you may end up with an entry in
-/// `epoll()` that cannot be removed (in the case of file descriptor-based
-/// sources), or an entry in some other data structure (eg. the `Timer`'s
-/// underlying heap structure). No unsoundness or undefined behaviour will
-/// result, but leaking file descriptors can result in errors or panics in a
-/// long running program.
+/// Not properly removing or replacing `TransientSource`s can cause spurious
+/// wakeups of the event loop, and in some cases can leak file descriptors or
+/// fail to free entries in Calloop's internal data structures. No unsoundness
+/// or undefined behaviour will result, but leaking file descriptors can result
+/// in errors or panics.
 ///
 /// If you want to remove a source before it returns `PostAction::Remove`, use
 /// the [`TransientSource::remove()`] method. If you want to replace a source
@@ -117,8 +115,7 @@
 ///
 /// If, instead, you directly assign a new source to the variable holding the
 /// `TransientSource`, the inner source will be dropped before it can be
-/// unregistered, resulting in a leak. For example, either of these assignments
-/// will cause a leak of the old source's registration:
+/// unregistered. For example:
 ///
 /// ```none,actually-rust-but-see-https://github.com/rust-lang/rust/issues/63193
 /// self.mpsc_receiver = Default::default();
--- a/src/sources/transient.rs
+++ b/src/sources/transient.rs
@@ -38,6 +38,9 @@
 /// let mpsc_receiver: TransientSource<Channel> = source.into();
 /// ```
 ///
+/// (If you want to start off with an empty `TransientSource`, you can just use
+/// `Default::default()` instead.)
+///
 /// `TransientSource` implements [`EventSource`](crate::EventSource) and passes
 /// through `process_events()` calls, so in the parent's `process_events()`
 /// implementation you can just do this:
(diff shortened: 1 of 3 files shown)


src/sources/transient.rs
```diff
@@ -103,53 +141,99 @@ pub enum TransientSource<T> {
...
+
+    /// Replace the currently wrapped source with the given one.  No more events
+    /// will be generated from the old source after this point. The old source
+    /// will not be dropped immediately, it will be kept so that it can be
+    /// deregistered.
+    ///
+    /// If this is called from outside of the event loop, you will need to wake
+    /// up the event loop for any changes to take place. If it is called from
+    /// within the event loop, the sources will be registered and unregistered
+    /// as needed after the current `process_events()` iteration.
```
vberger, line 225: ... assuming you correctly forward the `PostAction` output, if you are using `TransientSource` as an implementation detail of your own event source.

src/sources/transient.rs
```diff
@@ -93,8 +96,43 @@
...
+///
+/// ## Replacing or removing `TransientSource`s without leaking
+///
+/// It is possible to leak registration if you bypass the API of a
+/// `TransientSource`. "Leak registration" means you may end up with an entry in
+/// `epoll()` that cannot be removed (in the case of file descriptor-based
+/// sources), or an entry in some other data structure (eg. the `Timer`'s
+/// underlying heap structure). No unsoundness or undefined behaviour will
+/// result, but leaking file descriptors can result in errors or panics in a
+/// long running program.
```
vberger, line 108: Actually, I don't think it is that bad:

- If when dropping the event source you close the FD, when afaik epoll/kqueue will handle that gracefully, they don't internally keep it open and it won't cause errors.
- Leaking timers just means that the event loop will have a spurious wakeup at the time the timer expires.

Writing this, I'm actually wondering if there is actually much to worry about in the first place. :thinking:

detly: > If when dropping the event source you close the FD, when afaik epoll/kqueue will handle that gracefully, they don't internally keep it open and it won't cause errors.

I don't know about this. I checked this on an older version of Calloop (0.9.3) because that's what our code base was on. I added some logging statements for FDs getting added and removed from `epoll()`, and without these changes, the FDs just gradually increase as the edge case in the test is hit.

I **did not** do the same for timers, but it seems like the [binary heap in `TimerWheel`](https://github.com/Smithay/calloop/blob/613c839be75ba4f172d089bff0a94e5d3dfa3232/src/sources/timer.rs#L184) would just grow without actually unregistering the timers, even if they're dropped, no?

src/sources/transient.rs
detly: Oops, I put this in the type-level docs but not here. It should be in both places.

src/sources/transient.rs
vberger: > I don't know about this. I checked this on an older version of Calloop (0.9.3) because that's what our code base was on. I added some logging statements for FDs getting added and removed from epoll(), and without these changes, the FDs just gradually increase as the edge case in the test is hit.

I was pretty sure of the opposite, guess I was misremembering. You did close the FDs on dropping the sources?

> I did not do the same for timers, but it seems like the [binary heap in TimerWheel](https://github.com/Smithay/calloop/blob/613c839be75ba4f172d089bff0a94e5d3dfa3232/src/sources/timer.rs#L184) would just grow without actually unregistering the timers, even if they're dropped, no?

Yeah, but the timers would still fire and be consumed when they expire, it's just that no event source will actually react to them because their tokens no longer match anything. So unless you create a shit ton of timers into the far future and drop them without unregistering them, seems those would actually be short-lived leaks.

vberger: I mean, the epoll man page states this pretty clearly imo:

> Will closing a file descriptor cause it to be removed from all epoll interest lists?
>
> Yes, but be aware of the following point.  A file descriptor is a reference to an open file description (see open(2)). Whenever a file descriptor is duplicated via dup(2), dup2(2), fcntl(2) F_DUPFD, or fork(2), a new file descriptor referring to the same open file description is created. An open file description continues to exist until all file descriptors referring to it have been closed.
> 
> A file descriptor is removed from an interest list only after all the file descriptors referring to the underlying open file description have been closed.  This means that even after a file descriptor that is  part  of  an interest  list  has been closed, events may be reported for that file descriptor if other file descriptors referring to the same underlying file description remain open.  To prevent this happening, the file descriptor must be explicitly removed from the interest list (using epoll_ctl(2) EPOLL_CTL_DEL) before it is duplicated.  Alternatively, the application must ensure that all file descriptors are closed (which may be difficult if file  descriptors were duplicated behind the scenes by library functions that used dup(2) or fork(2)).

detly: > You did close the FDs on dropping the sources?

These are all executors or channels, which should handle that internally, no?

> Yeah, but the timers would still fire and be consumed when they expire, it's just that no event source will actually react to them because their tokens no longer match anything.

That's a relief then.

I will see if I can repro this on 0.10.x. The fact that the fds increased suggest that something is holding them open, somehow. If nothing else though, even if `epoll()` removes fds from its own side, the fds-as-indexes will remain on "our" side in the `sys::Poll` `tokens` list, won't they?

detly:
Taking the leak discussion out here. I can't yet repro the ever-increasing-fd leak I saw with the async executor; I'll have to trim down some application code until I can get it again.

However, I was able to show a couple of issues just using channels. All these branches have the same basic changes: some extra debugging in `src/sys/mod.rs` to show registration, and an example binary that drops and replaces a channel over and over again. In each branch you can run it with

```none
cargo run --example leak
```

Branches:

- https://github.com/detly/calloop/tree/source-leak-demonstration-0.9.3 - 0.9.3 with backported `TransientSource` (plus some cherry picks to make it easier; the important thing is it's pre-slotmap and pre-my-safety-changes)
- https://github.com/detly/calloop/tree/source-leak-demonstration-0.10.x - 0.10.1 pre this PR
- https://github.com/detly/calloop/tree/source-leak-demonstration-fixed - uses API from this PR

Effects:

- 0.9.3 - re-registers an FD without ever unregistering it.
- 0.10.x - panic due to changes I introduced, where we keep our map of FDs on the "Rust side" of FFI and panic if we get the same one twice. Arguably we could not panic and just warn.
- this PR - no panic, all FDs unregistered before being re-registered.

detly:
I feel like I'm losing it. I went back to my code from last week to try to un-fix what I fixed - no leak. Certainly I see sources not being unregistered, but I don't see the ever-increasing FD numbers I saw during my testing.

I think this is still a valid change - having FDs just quietly get closed but not updating our event loop structures is not ideal. But it's not the leak I thought it was.

vberger:
I agree that's still cleaner overall to properly unregister stuff like your PR does. I just think the doc comment can be adjusted to be less scary. What about something along the line of "Not properly unregistering the sources can cause spurious wakeups of the event loop, and *might* leak file descriptors in some cases." ?

vberger:
> Very reasonable point, except that with Calloop as it is now, it'll actually cause a panic. Do you want to relax the FD reregistration check? I'm not completely certain either way, myself.

Right, I meant "Not properly unregistering event source *before dropping them*". Removing an event source without unregistering it and then re-registering it again is definitely a wrong usage.

detly:
> > Very reasonable point, except that with Calloop as it is now, it'll actually cause a panic. Do you want to relax the FD reregistration check? I'm not completely certain either way, myself.
> 
> Right, I meant "Not properly unregistering event source _before dropping them_". Removing an event source without unregistering it and then re-registering it again is definitely a wrong usage.

I think you misunderstand, even if you don't re-use the same event source, as you point out the `CloseOnDrop` handler will close the fd, the platform will recycle the fd for the next eg. pipe or socket or file that gets used, and then the code in `src/sys/mod.rs` will see a duplicate fd. Hence the panic.

detly:
Made the wording a bit less dire, anyway, but still mentioned the possibility of panicking.

src/sources/transient.rs
```diff
@@ -267,11 +369,12 @@ impl<T: crate::EventSource> crate::EventSource for TransientSource<T> {
 mod tests {
```
i509VCB, line 369: A bit of a nitpick, but the tests module is huge after this pull request, it's nearly 3x the size of the rest of the file.

Would it make sense to move the tests to their own file `src/sources/transient/tests.rs`?

detly: Fine by me. Arguably these are integration tests, not unit tests anyway. I'd happily put them in `/tests/`, but then it'd be a consistency issue with all the other not-actually-unit-tests.

vberger: That's right that many of the tests in calloop can probably be moved to `/tests/`, but I suggest we leave that for another PR.

vberger (approved):

PR #108, opened 2022-07-26
Use pipe fds in insert_source_no_interest test instead of stdin.

### LINKED ISSUE #96: insert_source_no_interest fails if stdin is redirected
I ran into this while trying to package calloop for Debian.  The automated test environment runs with stdin redirected from `/dev/null`, causing the test to fail. 

I added the following diff to shed light on the specific error:

```diff
diff --git i/src/loop_logic.rs w/src/loop_logic.rs
index bb3a6d3..49e53f1 100644
--- i/src/loop_logic.rs
+++ w/src/loop_logic.rs
@@ -617,7 +617,7 @@ mod tests {
             crate::sources::generic::Generic::from_fd(0, Interest::EMPTY, Mode::Level),
             |_, _, _| Ok(PostAction::Continue),
         );
-        assert!(ret.is_ok());
+        assert!(ret.is_ok(), "{:?}", ret);
     }
 
     #[test]
```

```
$ cargo test -- no_interest </dev/null
    Finished test [unoptimized + debuginfo] target(s) in 0.02s
     Running unittests (target/debug/deps/calloop-035331c4b39aed70)

running 1 test
test loop_logic::tests::insert_source_no_interest ... FAILED

failures:

---- loop_logic::tests::insert_source_no_interest stdout ----
thread 'loop_logic::tests::insert_source_no_interest' panicked at 'Err(Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" })', src/loop_logic.rs:620:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    loop_logic::tests::insert_source_no_interest

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 26 filtered out; finished in 0.00s
```

detly:
Fixes #96 where this test would fail when stdin is closed (eg. by redirecting
from /dev/null).

*(I copied the lowest-common-denomimator from the ping sources, so it emulated the existing behaviour where we just used a raw fd, but maybe it's worth just using our own Ping source?)*

### DIFF
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -568,12 +568,28 @@ mod tests {
 
     #[test]
     fn insert_source_no_interest() {
+        use nix::unistd::{close, pipe};
+
+        // Create a pipe to get an arbitrary fd.
+        let (read, write) = pipe().unwrap();
+        // We don't need the write end.
+        close(write).unwrap();
+
+        let source = crate::sources::generic::Generic::new(read, Interest::EMPTY, Mode::Level);
+        let dispatcher = Dispatcher::new(source, |_, _, _| Ok(PostAction::Continue));
+
         let event_loop = EventLoop::<()>::try_new().unwrap();
-        let ret = event_loop.handle().insert_source(
-            crate::sources::generic::Generic::new(0, Interest::EMPTY, Mode::Level),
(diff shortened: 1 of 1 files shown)


vberger:
Thanks!

PR #110, opened 2022-08-25
Fix calloop double borrows

### LINKED ISSUE #109: Double borrows when registering sources and idle callbacks
Inside of `pre_run` and `post_run`, registering an event source will cause a double borrow.

This is also possible inside of the callback provided to `insert_idle` since registering an idle callback inside of an idle callback will double borrow.

I reproduced the double borrow with the following code:
```rust
use calloop::{
    ping::PingSource,
    EventLoop, EventSource, LoopHandle, Poll, PostAction, Readiness, Token, TokenFactory,
};

struct State {
    loop_handle: LoopHandle<'static, Self>,
}

#[test]
fn insert_in_dispatch() {
    let mut event_loop = EventLoop::try_new().unwrap();

    let mut state = State {
        loop_handle: handle.clone(),
    };

    let timer = OneshotRemove::new();

    handle.insert_source(timer, handle_source).unwrap();

    event_loop.run(None, &mut state, |_| {}).unwrap();
}

fn handle_source(_: (), _: &mut (), state: &mut State) {
    state
        .loop_handle
        .insert_source(OneshotRemove::new(), handle_source)
        .unwrap();
}

struct OneshotRemove {
    source: PingSource,
}

impl OneshotRemove {
    pub fn new() -> Self {
        let (ping, source) = calloop::ping::make_ping().unwrap();
        let source = Self { source };

        // Immediately ping the source to dispatch on register
        ping.ping();
        source
    }
}

impl EventSource for OneshotRemove {
    type Event = ();
    type Metadata = ();
    type Ret = ();
    type Error = calloop::Error;

    fn pre_run<F>(&mut self, mut callback: F) -> calloop::Result<()>
    where
        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
    {
        callback((), &mut ());
        Ok(())
    }

    fn process_events<F>(
        &mut self,
        readiness: Readiness,
        token: Token,
        _callback: F,
    ) -> Result<PostAction, Self::Error>
    where
        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
    {
        self.source
            .process_events(readiness, token, |_, _| {})
            .unwrap();

        Ok(PostAction::Remove)
    }

    fn register(
        &mut self,
        poll: &mut Poll,
        token_factory: &mut TokenFactory,
    ) -> calloop::Result<()> {
        self.source.register(poll, token_factory)
    }

    fn reregister(
        &mut self,
        poll: &mut Poll,
        token_factory: &mut TokenFactory,
    ) -> calloop::Result<()> {
        self.source.reregister(poll, token_factory)
    }

    fn unregister(&mut self, poll: &mut Poll) -> calloop::Result<()> {
        self.source.unregister(poll)
    }
}
```

The following backtrace is produced:
https://gist.github.com/i509VCB/0b27d3a5eb57ae7594243cd5b4833cf7

i509VCB:
Fixes #109

### DIFF
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -1,11 +1,11 @@
 use std::cell::{Cell, RefCell};
 use std::fmt::Debug;
-use std::io;
 use std::os::unix::io::AsRawFd;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+use std::{io, mem};
 
 use slotmap::SlotMap;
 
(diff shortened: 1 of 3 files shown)


chrisduerr:
I can also confirm that this fixes my issue, so should be good to go hopefully.

src/loop_logic.rs
```diff
@@ -358,23 +358,58 @@ impl<'l, Data> EventLoop<'l, Data> {
...
 
     fn dispatch_idles(&mut self, data: &mut Data) {
-        let idles = ::std::mem::take(&mut *self.handle.inner.idles.borrow_mut());
+        // Collect all idle callbacks to prevent RefCell double borrows.
+        let idles = {
+            let mut idles = self.handle.inner.idles.borrow_mut();
+
+            // Idle callbacks are oneshot, replace the Vec
+            mem::take(&mut *idles)
+        };
```
vberger, line 369: Can't we just do this ?

```rust
let idles = mem::take(&mut *self.handle.inner.idles.borrow_mut());
```

vberger (commented):
LGTM overall, this is the correct fix. :+1: 

There are a few places where you've enclosed the expressions in `{ ... }`, are those really necessary? They seem redundant.

src/loop_logic.rs
i509VCB: The intention was to be explicit with ending the scope of the refcell borrows.

I can probably move to using just using drop(cell)

vberger: IMO the that that we're collecting the callbacks out, with the comment explaining why is already explicit enough. Those redundant scopes are just making the whole code heavier.

kchibisov:
Might be related to this as well https://github.com/alacritty/alacritty/issues/6487

kchibisov:
@i509VCB If you don't feel like finishing this PR(some codestyle suggestions from @elinorbgr) , let me know, so I can finish it, since right now it's causing crashes downstream which is frustrating.

elinorbgr:
Replaced by #115

PR #120, opened 2022-12-18
Add a "block_on" function for executing futures

notgull:
This PR adds a "block on" function to the `EventLoop` that serves a similar purpose as the [`block_on()`](https://docs.rs/async-io/latest/async_io/fn.block_on.html) function from `async_io`, where the event loop is polled during periods where the future would normally be blocked. I gated it behind a `block_on` feature, please let me know if it would be better to not gate it.

### DIFF
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -7,6 +7,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+#[cfg(feature = "block_on")]
+use std::future::Future;
+
 use slotmap::SlotMap;
 
 use crate::sources::{Dispatcher, EventSource, Idle, IdleDispatcher};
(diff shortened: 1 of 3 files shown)


src/loop_logic.rs
```diff
@@ -892,6 +990,62 @@ mod tests {
...
+
+        let mut data = 22;
+        let timeout = {
+            let data = &mut data;
+
+            async move {
+                async_io::Timer::after(Duration::from_secs(1)).await;
+                *data = 32;
+                11
+            }
```
vberger, line 1006: calloop has its own timer futures (`calloop::timer::TimeoutFuture`), we can use them instead of pulling async-io as a dev dependency. Furthermore this would ensure that the added function is indeed able to work with calloop-driven futures.

vberger:
Thanks for your PR. It looks like a pretty good addition! Could you add a changelog entry for this addition please?

notgull:
@vberger Is there anything blocking this from being merged?

vberger:
Apart for my own availability, no. Thanks!

PR #122, opened 2022-12-22
Bump `nix` to 0.26.0

ids1024:


vberger:
Ah, looks like this bumps the MSRV?

ids1024:
Hm, yeah, looks like it.

ids1024:
Seems 1.56 is the oldest version that supports this, so might as well bump the MSRV to that. Wayland-rs is using 1.59, so I imagine there aren't any issues with this.

PR #124, opened 2023-04-09
Replace futures-util with async-task

notgull:
WIP


elinorbgr:
Looks like your PR changed some behavior, as a test is no longer passing, and it does not look like a fluke.

notgull:
> Looks like your PR changed some behavior, as a test is no longer passing, and it does not look like a fluke.

Yeah, I misread how the channels work and made it so it only drained 1 `Runnable` instead of a chunk of `Runnable`s, which messed up the dispatching tests. This should be fixed now.

src/sources/futures.rs
```diff
@@ -36,16 +43,69 @@ use crate::{
...
+    active: RefCell<Option<Slab<Active<T>>>>,
+}
+
+/// Send a future to an executor.
+///
+/// This needs to be thread-safe, as it is called from a `Waker` that may be on a different thread.
+#[derive(Debug)]
+struct Sender {
+    /// The sender used to send runnables to the executor.
+    sender: Mutex<mpsc::Sender<Runnable<usize>>>,
```
elinorbgr, line 81: Just so that we don't have to wonder about why this mutex is here when we reread this later.

```suggestion
    /// The sender used to send runnables to the executor.
    ///
    /// (`mpsc::Sender` is `!Sync`)
    sender: Mutex<mpsc::Sender<Runnable<usize>>>,
```

src/sources/futures.rs
```diff
@@ -36,16 +43,69 @@ use crate::{
...
+    /// The incoming queue of runnables to be executed.
+    incoming: mpsc::Receiver<Runnable<usize>>,
+
+    /// The sender corresponding to `incoming`.
+    sender: Arc<Sender>,
+
+    /// The list of currently active tasks.
+    ///
+    /// This is set to `None` when the executor is destroyed.
+    active: RefCell<Option<Slab<Active<T>>>>,
```
elinorbgr, line 72: I'd prefer to rename this field into `active_tasks` (as well as the locals borrowing it in the code), to ease with readability.

src/sources/futures.rs
```diff
@@ -56,43 +116,187 @@ impl<T> Scheduler<T> {
...
-#[error("the executor was destroyed")]
-pub struct ExecutorDestroyed;
+impl Sender {
+    /// Send a runnable to the executor.
+    fn send(&self, runnable: Runnable<usize>) {
+        // Send on the channel.
+        if let Err(e) = self
+            .sender
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
```
elinorbgr, line 229: Would be better to have a comment explaining why ignoring the poisoning error is fine, if it is. And if not, to handle it properly.

src/sources/futures.rs
```diff
@@ -56,43 +116,187 @@ impl<T> Scheduler<T> {
...
+        {
+            // Make sure the runnable is never dropped.
+            std::mem::forget(e);
 
-#[derive(Debug)]
-struct ExecWaker {
-    ping: Ping,
+            // This shouldn't be able to happen since all of the tasks are destroyed before the
+            // executor is. This indicates a critical soundness bug.
+            std::process::abort();
```
elinorbgr, line 237: Is there a reason to `abort()` instead of panicking?

src/sources/futures.rs
```diff
@@ -56,43 +116,187 @@ impl<T> Scheduler<T> {
...
+    fn send(&self, runnable: Runnable<usize>) {
+        // Send on the channel.
+        if let Err(e) = self
+            .sender
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+            .send(runnable)
+        {
+            // Make sure the runnable is never dropped.
+            std::mem::forget(e);
```
elinorbgr, line 233: Why?

src/sources/futures.rs
```diff
@@ -56,43 +116,187 @@ impl<T> Scheduler<T> {
...
-    fn wake_by_ref(arc_self: &Arc<ExecWaker>) {
-        arc_self.ping.ping();
+impl<T> Drop for Executor<T> {
+    fn drop(&mut self) {
+        let active = self.state.active.borrow_mut().take().unwrap();
+
+        // Wake all of the active tasks in order to destroy their runnables.
+        for (_, task) in active {
+            // Don't let a panicking waker blow everything up.
+            std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
elinorbgr, line 257: Why not? And how are we sure that this `AssertUnwindSafe` is indeed correct?

src/sources/futures.rs
notgull: This is a safe guard in the event that something I've overlooked causes unsound behavior. In this case, the Sender shouldn't fail, but if it does, it means that something has gone terribly unsound. In this case aborting is the only option, as panicking may let the program continue with unsound behavior.

src/sources/futures.rs
notgull: I've figured out a way around the `AssertUnwindSafe`.

src/sources/futures.rs
elinorbgr: I see, but in that case we should at least print *something* to the user, that's why I believe panicking would be preferable: it gives relevant information to debug & understand what occurred.

Indeed the user may technically catch the panic and continue running, but it is widely admitted in the community that it's the responsibility of the one catching the panic to ensure that what they are doing is still safe. So I'd prefer a panic with a relevant error message: that sounds like a place for using `unreachable!()`.

src/sources/futures.rs
elinorbgr: There is still the initial question: why catch the panic here?

notgull: Let's say a future panics. While the stack unwinds, the executor is dropped, which wakes up all of these wakers so that the tasks are all queued for the last time.

Since the future panicked, it's likely that something else was put into an invalid state and it's not out of the question that the waker will panic too. However, since the thread is currently panicking, a waker panic will be upgraded into an abort.

In the interest of not aborting, the waker panics here are just ignored.

src/sources/futures.rs
notgull: Sounds good; I panic instead.

src/sources/futures.rs
elinorbgr: Could you add this explanation as a comment alongside this code then?

elinorbgr (approved):
Alright, LGTM, thanks!

PR #131, opened 2023-04-29
Port to rustix from nix

notgull:
Ports parts of this crate from nix to rustix, which uses I/O safety by default and allows for the inlining of syscalls.

WIP because there are some features we need that `rustix` does not support yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,7 +86,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ['1.56.0', 'stable', 'beta']
+        rust: ['1.63.0', 'stable', 'beta']
 
     runs-on: 'ubuntu-latest'
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,12 @@ jobs:
           command: test
           args: --features "block_on executor"
 
+      - name: Run tests with signals
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features "block_on executor signals"
+
       - name: Run book tests
         uses: actions-rs/cargo@v1
         if: ${{ matrix.rust == 'stable' || matrix.rust == 'beta' }}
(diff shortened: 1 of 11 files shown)


notgull:
Unfortunately it looks like you really do need to talk to `libc` for signals. So I'll keep `nix` around for that, but make it optional, as it looks like none of `calloop`'s dependencies explicitly use it.

ids1024:
Looking at `signals.rs`, I guess that's for pthreads stuff?

Looks like `thread_block` and `thread_unblock` in nix call `pthread_sigmask`, and looking at musl, that's a wrapper around the syscall `SYS_rt_sigprocmask`. So if the few other calls there are similar, it should be possible to do with directly syscalls on Linux (and library calls elsewhere). Other pthreads APIs may not be as thin a wrapper over system calls.

notgull:
IIRC the "wrapper" in Musl is a lot less thin than you would expect. See [this issue](https://github.com/bytecodealliance/rustix/issues/628) in `rustix` for more information, as well as the [undocumented `sigprocmask` function](https://github.com/bytecodealliance/rustix/blob/ebcfe14593d0448aa91cd855d67003d8db34a232/src/runtime.rs#L367-L385) currently in `rustix::runtime`. So we probably shouldn't do that.

notgull:
Looks like CI is passing. @elinorbgr What are your thoughts?

.github/workflows/ci.yml
```diff
@@ -137,6 +137,12 @@ jobs:
           command: test
           args: --features "block_on executor"
 
+      - name: Run tests with signals
+        uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features "block_on executor signals"
+
```
elinorbgr, line 145: Why do this second test step in rather than just enabling the `signals` feature on the previous entry?

elinorbgr (commented):
Looks like a pretty straightforward PR, thanks!

Just one small point regarding the CI, otherwise LGTM :+1:

.github/workflows/ci.yml
notgull: In order to test the absence of `nix`, which was an issue while I was testing it.

It can be an issue sometimes where  some permutations of features have compile errors without the maintainer knowing. It might be a good idea to use [`cargo-hack`](https://github.com/taiki-e/cargo-hack) in the CI for this crate.

PR #141, opened 2023-07-11
Update for new polling breaking changes

notgull:
For now, this exists as a testing ground in order to test https://github.com/smol-rs/polling/pull/123 before we release it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,3 +48,6 @@ rustdoc-args = ["--cfg", "docsrs"]
 [[test]]
 name = "signals"
 harness = false
+
+[patch.crates-io]
+polling = { git = "https://github.com/smol-rs/polling.git", branch = "notgull/unsafe2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ io-lifetimes = "1.0.3"
 log = "0.4"
 nix = { version = "0.26", default-features = false, features = ["signal"], optional = true }
 pin-utils = { version = "0.1.0", optional = true }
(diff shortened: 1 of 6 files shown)


ids1024:
So, this makes `Generic` safe by having `Drop` and `unwrap` automatically unregister it from the event loop?

Interesting approach. Normally relying on `Drop` for safety properties is unsound, with `mem::forget` or reference cycles, but in this particular instance it's fine since that will also leak the fd. (In other words, I tried to break this and get it to poll an invalid fd using `Generic`, but I couldn't. So it looks sound.)

But this does mean that if a user of `calloop` implements `EventSource` to call `Poll::register` etc. directly, it should also implement `Drop` like this to ensure the same safety properties?

notgull:
The only thing that matters is that the source isn't dropped while registered in a `Poller`. If anyone were to implement a similar abstraction to this, they would need to either follow these dropping rules or otherwise make sure that `drop()` isn't called before `deregister()`.

notgull:
I've updated this for the new `polling` release.

notgull:
Ugh, looks like Github Actions is down.

elinorbgr:
Alright, the CI is back. Also, even if git does not see any conflict, could you rebase on master for good measure?

elinorbgr:
Okay, this looks good overall, but could you give more details about the `NoIoDrop` type that you introduced? Mostly what is its role and why is it needed? If possible as code/docs comments around this type, to make future maintainability easier.

elinorbgr (approved):
Perfect, thanks!

Issue #127, opened 2023-04-17
Allow event source to mark itself as being ready from the `pre_run` or similar


kchibisov:
Some event sources could be read be read outside of the calloop reading.

An example of such source could be a `WaylandSource`. The wayland display could be read
by other libraries, like mesa, to check that you have events, the `wl_display_prepare_read`
is used. Such issue created a [workaround](https://github.com/rust-windowing/winit/blob/60e91b187a9cc68c38dbe2cd9c83ebf2ee2f723e/src/platform_impl/linux/wayland/event_loop/mod.rs#L265) inside the `winit` library.

One possible solution would be to add a way for event source to set the indicator about instant
wake ups, so calloop when trying to poll on them again will do instant wakeup instead and call the
event/read part.

cc @rib

rib:
Yeah, so after I had initially convinced myself that `calloop` was using "prepare_read" and therefore we don't need that workaround in Winit - I think now that calloop isn't currently following the standard protocol that's required to allow multiple threads to safely access a shared Wayland socket.

E.g. here's a C reference for how the prepare_read protocol can be followed to safely support multi-threaded access without deadlocks: https://wayland.freedesktop.org/docs/html/apb.html#Client-classwl__display_1a40039c1169b153269a3dc0796a54ddb0

Also here's a reference for implementing a similar kind of source for the glib main loop: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdkeventsource.c#L63

It looks like the logic has changed between smithay-client-toolkit-0.16 and master but in either case it doesn't seem to follow the ordering of the reference C code linked above.

elinorbgr:
Coming back to this, I believe the proper way to fix this is to replace the pre_run API with something with the following semantics:

- A method that says to the event source "I'm going to sleep", and from with the source can answer "actually I'm already ready"
- A method that notifies the source that we are done sleeping (be it to process the events, or returning from the event loop), and this method should also tell the source whether it has been woken up or not, so that it can do some preprocessing if necessary (but *not* invoke user code from it).

This should allow to fix the Wayland issues, which I believe come from the fact that we should not keep the source in a "preparing to read events" state while we are doing something else than sleeping on the socket (for example while we are running user callbacks).

kchibisov:
>A method that says to the event source "I'm going to sleep", and from with the source can answer "actually I'm already ready"

[...]

PR #144, opened 2023-08-26
Add opt-in handlers for before sleeping and before handling events

DJMcNab:
Solves #127. Designed to work with https://github.com/Smithay/wayland-rs/pull/650 and to be used to fix [#glazier > Freeze with Wayland egl](https://xi.zulipchat.com/#narrow/stream/351333-glazier/topic/Freeze.20with.20Wayland.20egl)


src/sources/mod.rs
```diff
@@ -356,15 +372,57 @@ pub(crate) trait EventDispatcher<Data> {
         data: &mut Data,
     ) -> crate::Result<PostAction>;
 
-    fn register(&self, poll: &mut Poll, token_factory: &mut TokenFactory) -> crate::Result<()>;
+    fn register(
+        &self,
+        poll: &mut Poll,
+        additional_lifetime_register: AdditionalLifetimeEventsRegister<'_>,
```
kchibisov, line 378: What's the reason to pass this thing around instead of storing sources with lifetimes and sources without in separate buckets and simply pick based on the assoc constant? And have a way to iterate all the sources which simply takes both backets and `chain`s them?

src/sources/mod.rs
```diff
@@ -157,25 +157,28 @@ pub trait EventSource {
...
-    where
-        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
-    {
-        Ok(())
-    }
+    /// Use this to perform a cleanup before event handlers with arbitrary
+    /// code may run. This could be used to drop a lock obtained in
+    /// [`EventSource::before_will_sleep`]
+    #[allow(unused_variables)]
+    fn before_handle_events(&mut self, was_awoken: bool) {}
```
kchibisov, line 181: That looks fine.

src/sources/mod.rs
DJMcNab: > And have a way to iterate all the sources which simply takes both backets and chains them?

I thought the operation which we perform isn't iterating, it's `.get`. If that's right, then the method would need to do two get based lookups for each event, which is potentially annoying. Additionally, due to the use of the slab datastructure, this would double the memory usage of the event loop, which isn't ideal.

src/loop_logic.rs
```diff
@@ -56,6 +58,7 @@ pub struct RegistrationToken {
 pub(crate) struct LoopInner<'l, Data> {
     pub(crate) poll: RefCell<Poll>,
     pub(crate) sources: RefCell<Slab<Rc<dyn EventDispatcher<Data> + 'l>>>,
+    pub(crate) sources_with_additional_lifetime_events: AdditionalLifetimeEventsSet,
```
kchibisov, line 61: Could we just duplicate the way sources are defined instead of rolling new abstraction? And create an iterator which simply chains `both` when needed?

src/sources/mod.rs
kchibisov: I mean, the buckets won't intersect at all. It's not like `Slab` will pre-allocate lots of storage just because you start using 2 of them?

You also maintain `HashSet` so it's it's hard to judge memory usage until you profile both approaches.

elinorbgr: Note, it is possible to iterate over the Slab contents, but this is discouraged because not efficient: https://docs.rs/slab/latest/slab/struct.Slab.html#method.iter

DJMcNab: I guess we could lose another bit in `Token` to storing whether the source needs additional lifecycle events. I'm not convinced that is better, though.

src/sources/mod.rs
```diff
@@ -85,10 +85,17 @@ impl BitOrAssign for PostAction {
...
-/// default. Depending on the underlying events, `process_events` may be invoked once,
-/// several times, or none at all between `pre_run` and `post_run` are called, but when
-/// it is invoked, it'll always be between those two.
+/// after that), you can override [`NEEDS_EXTRA_LIFECYCLE_EVENTS`] to `true`.
+/// For all sources for which that constant is `true`, the methods [`before_will_sleep`] and
+/// [`before_handle_events`] will be called.
+/// [`before_will_sleep`] is called before the polling system performs a poll operation.
+/// [`before_handle_events`] is called before any process_events methods have been called.
+/// This means that during `process_events` you can assume that all cleanup has occured on
+/// all sources.
```
elinorbgr, line 94: I'd prefer if this comment also gave some notice about how to handle this when building composite sources, so set the constant to `true` if at least one of the sub-sources need it, and forward the calls to the two methods to the sub-sources that need it.

Also, writing this, I realize that `before_will_sleep` would need to return a `Vec<(Readiness, Token))>` (or something similar), rather than just an option, to cover the case when a composite source has more than one sub-source that needs this...

src/sources/mod.rs
```diff
@@ -157,25 +164,28 @@ pub trait EventSource {
...
-        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
-    {
-        Ok(())
-    }
-
-    /// Notification that the current polling session ended
+    /// Use this to perform operations which must be done before polling,
+    /// but which may conflict with other event handlers. For example,
+    /// if polling requires a lock to be taken.
+    fn before_will_sleep(&mut self) -> crate::Result<Option<(Readiness, Token)>> {
```
elinorbgr, line 178: What about naming this `before_sleep` instead? This matches better the naming pattern of the other method.

src/sources/mod.rs
elinorbgr: Is this something we care about supporting, or do we declare this "too niche"?

src/sources/mod.rs
elinorbgr: So, as discussed on Matrix, the main point in favor of the current structure is that it allows to track when sources are disabled and reenabled, to avoid invoking `before_sleep()` on a disabled source. That seems like an important point to me.

elinorbgr:
Also, it looks like the `signals` test is failing? Is it because of the nix update?

src/sources/mod.rs
DJMcNab: We could pass in a wrapper around `&mut Vec<(Readiness, Token)>`, if we wanted. But I think it's not worth adding preemptively.

Cargo.toml
```diff
@@ -6,23 +6,28 @@ documentation = "https://docs.rs/calloop/"
...
 
 [badges]
 codecov = { repository = "Smithay/calloop" }
 
 [dependencies]
-bitflags = "1.2"
 io-lifetimes = "1.0.3"
 log = "0.4"
-nix = { version = "0.26", default-features = false, features = ["event", "fs", "signal", "socket", "time"] }
+nix = { version = ">=0.26.0, <0.26.3", default-features = false, features = [
```
elinorbgr, line 24: 0.26.3 has been yanked, so I guess we can restore this to just `0.26`

src/sources/mod.rs
```diff
@@ -356,15 +379,57 @@ pub(crate) trait EventDispatcher<Data> {
...
+impl<'a> AdditionalLifetimeEventsRegister<'a> {
+    fn register(&mut self) {
+        self.set.values.borrow_mut().push((self.token, false))
+    }
 
-    fn post_run(&self, data: &mut Data) -> crate::Result<()>;
+    fn unregister(&mut self) {
+        self.set.values.borrow_mut().retain(|it| it.0 != self.token)
+    }
 }
```
elinorbgr, line 433: Could add a few comments over these types and methods, to keep a trace of what they are for and how they are used? That'd make future maintenance of this code easier.

elinorbgr (changes requested):
There are naming inconsistencies in the code, sometimes things are named "lifecycle events" and sometimes "lifetime events", I think we should use "lifecycle events" everywhere.

Also, could you rebase this PR on master? Otherwise, lgtm :+1:

kchibisov (changes requested):
Can't we at least do something like that? it should at least make it more readable. The main pain point is still `Rc` stuff. Also, I'm not sure what this all logic wrt Registration token, TokenFactory tries to do, because it's not consistent with what it does. Like it seems that sometimes sub sources are being passed, sometimes real sources, it's just all over the place, but I don't want to go into details with it.

My patch is likely broken with regards to those tokens, but generally it's just more consistent with what already is. I don't disagree with the idea to cache tokens to improve lookups, it' just I'd rather follow the current abstractions.

```diff
diff --git a/src/io.rs b/src/io.rs
index ffbb857..980cd59 100644
--- a/src/io.rs
+++ b/src/io.rs
@@ -18,7 +18,7 @@ use rustix::fs::{fcntl_getfl, fcntl_setfl, OFlags};
 #[cfg(feature = "futures-io")]
 use futures_io::{AsyncRead, AsyncWrite, IoSlice, IoSliceMut};
 
-use crate::AdditionalLifetimeEventsRegister;
+use crate::AdditionalLifecycleEventsSet;
 use crate::{
     loop_logic::{LoopInner, MAX_SOURCES_MASK},
     sources::EventDispatcher,
@@ -241,7 +241,7 @@ impl<Data> EventDispatcher<Data> for RefCell<IoDispatcher> {
     fn register(
         &self,
         _: &mut Poll,
-        _: AdditionalLifetimeEventsRegister<'_>,
+        _: &mut AdditionalLifecycleEventsSet,
         _: &mut TokenFactory,
     ) -> crate::Result<()> {
         // registration is handled by IoLoopInner
@@ -251,7 +251,7 @@ impl<Data> EventDispatcher<Data> for RefCell<IoDispatcher> {
     fn reregister(
         &self,
         _: &mut Poll,
-        _: AdditionalLifetimeEventsRegister<'_>,
+        _: &mut AdditionalLifecycleEventsSet,
         _: &mut TokenFactory,
     ) -> crate::Result<bool> {
         // registration is handled by IoLoopInner
@@ -261,7 +261,8 @@ impl<Data> EventDispatcher<Data> for RefCell<IoDispatcher> {
     fn unregister(
         &self,
         poll: &mut Poll,
-        _: AdditionalLifetimeEventsRegister<'_>,
+        _: &mut AdditionalLifecycleEventsSet,
+        _: TokenFactory,
     ) -> crate::Result<bool> {
         let disp = self.borrow();
         if disp.is_registered {
diff --git a/src/loop_logic.rs b/src/loop_logic.rs
index a0ab8e5..661e58d 100644
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -15,7 +15,7 @@ use slab::Slab;
 use crate::sources::{Dispatcher, EventSource, Idle, IdleDispatcher};
 use crate::sys::{Notifier, PollEvent};
 use crate::{
-    AdditionalLifetimeEventsSet, EventDispatcher, InsertError, Poll, PostAction, TokenFactory,
+    AdditionalLifecycleEventsSet, EventDispatcher, InsertError, Poll, PostAction, TokenFactory,
 };
 
 type IdleCallback<'i, Data> = Rc<RefCell<dyn IdleDispatcher<Data> + 'i>>;
@@ -58,7 +58,7 @@ pub struct RegistrationToken {
 pub(crate) struct LoopInner<'l, Data> {
     pub(crate) poll: RefCell<Poll>,
     pub(crate) sources: RefCell<Slab<Rc<dyn EventDispatcher<Data> + 'l>>>,
-    pub(crate) sources_with_additional_lifecycle_events: AdditionalLifetimeEventsSet,
+    pub(crate) sources_with_additional_lifecycle_events: RefCell<AdditionalLifecycleEventsSet>,
     idles: RefCell<Vec<IdleCallback<'l, Data>>>,
     pending_action: Cell<PostAction>,
 }
@@ -140,9 +140,10 @@ impl<'l, Data> LoopHandle<'l, Data> {
         let key = sources.insert(dispatcher.clone_as_event_dispatcher());
         let ret = sources.get(key).unwrap().register(
             &mut poll,
-            self.inner
+            &mut self
+                .inner
                 .sources_with_additional_lifecycle_events
-                .create_register_for_token(RegistrationToken { key }),
+                .borrow_mut(),
             &mut TokenFactory::new(key),
         );
 
@@ -178,9 +179,10 @@ impl<'l, Data> LoopHandle<'l, Data> {
         if let Some(source) = self.inner.sources.borrow().get(token.key) {
             source.register(
                 &mut self.inner.poll.borrow_mut(),
-                self.inner
+                &mut self
+                    .inner
                     .sources_with_additional_lifecycle_events
-                    .create_register_for_token(*token),
+                    .borrow_mut(),
                 &mut TokenFactory::new(token.key),
             )?;
         }
@@ -195,9 +197,10 @@ impl<'l, Data> LoopHandle<'l, Data> {
         if let Some(source) = self.inner.sources.borrow().get(token.key) {
             if !source.reregister(
                 &mut self.inner.poll.borrow_mut(),
-                self.inner
+                &mut self
+                    .inner
                     .sources_with_additional_lifecycle_events
-                    .create_register_for_token(*token),
+                    .borrow_mut(),
                 &mut TokenFactory::new(token.key),
             )? {
                 // we are in a callback, store for later processing
@@ -214,9 +217,11 @@ impl<'l, Data> LoopHandle<'l, Data> {
         if let Some(source) = self.inner.sources.borrow().get(token.key) {
             if !source.unregister(
                 &mut self.inner.poll.borrow_mut(),
-                self.inner
+                &mut self
+                    .inner
                     .sources_with_additional_lifecycle_events
-                    .create_register_for_token(*token),
+                    .borrow_mut(),
+                TokenFactory::new(token.key),
             )? {
                 // we are in a callback, store for later processing
                 self.inner.pending_action.set(PostAction::Disable);
@@ -230,9 +235,11 @@ impl<'l, Data> LoopHandle<'l, Data> {
         if let Some(source) = self.inner.sources.borrow_mut().try_remove(token.key) {
             if let Err(e) = source.unregister(
                 &mut self.inner.poll.borrow_mut(),
-                self.inner
+                &mut self
+                    .inner
                     .sources_with_additional_lifecycle_events
-                    .create_register_for_token(token),
+                    .borrow_mut(),
+                TokenFactory::new(token.key),
             ) {
                 log::warn!(
                     "[calloop] Failed to unregister source from the polling system: {:?}",
@@ -320,16 +327,17 @@ impl<'l, Data> EventLoop<'l, Data> {
     ) -> crate::Result<()> {
         let now = Instant::now();
         {
-            let mut extra_lifecycle_sources = self
+            let extra_lifecycle_sources = &mut self
                 .handle
                 .inner
                 .sources_with_additional_lifecycle_events
-                .values
-                .borrow_mut();
+                .borrow_mut()
+                .values;
+
             let sources = &self.handle.inner.sources.borrow();
-            for (source, has_event) in &mut *extra_lifecycle_sources {
+            for (source_key, has_event) in &mut *extra_lifecycle_sources {
                 *has_event = false;
-                if let Some(disp) = sources.get(source.key) {
+                if let Some(disp) = sources.get(*source_key) {
                     if let Some((readiness, token)) = disp.before_sleep()? {
                         // Wake up instantly after polling if we recieved an event
                         timeout = Some(Duration::ZERO);
@@ -363,21 +371,21 @@ impl<'l, Data> EventLoop<'l, Data> {
             }
         };
         {
-            let mut extra_lifecycle_sources = self
+            let extra_lifecycle_sources = &mut self
                 .handle
                 .inner
                 .sources_with_additional_lifecycle_events
-                .values
-                .borrow_mut();
+                .borrow_mut()
+                .values;
             if !extra_lifecycle_sources.is_empty() {
-                for (source, has_event) in &mut *extra_lifecycle_sources {
+                for (source_key, has_event) in &mut *extra_lifecycle_sources {
                     for event in &events {
-                        *has_event |= (event.token.key & MAX_SOURCES_MASK) == source.key;
+                        *has_event |= (event.token.key & MAX_SOURCES_MASK) == *source_key;
                     }
                 }
             }
-            for (source, has_event) in &*extra_lifecycle_sources {
-                if let Some(disp) = self.handle.inner.sources.borrow().get(source.key) {
+            for (source_key, has_event) in &*extra_lifecycle_sources {
+                if let Some(disp) = self.handle.inner.sources.borrow().get(*source_key) {
                     disp.before_handle_events(*has_event);
                 } else {
                     unreachable!()
@@ -414,24 +422,23 @@ impl<'l, Data> EventLoop<'l, Data> {
                     PostAction::Reregister => {
                         disp.reregister(
                             &mut self.handle.inner.poll.borrow_mut(),
-                            self.handle
+                            &mut self
+                                .handle
                                 .inner
                                 .sources_with_additional_lifecycle_events
-                                .create_register_for_token(RegistrationToken {
-                                    key: registroken_token,
-                                }),
+                                .borrow_mut(),
                             &mut TokenFactory::new(event.token.key),
                         )?;
                     }
                     PostAction::Disable => {
                         disp.unregister(
                             &mut self.handle.inner.poll.borrow_mut(),
-                            self.handle
+                            &mut self
+                                .handle
                                 .inner
                                 .sources_with_additional_lifecycle_events
-                                .create_register_for_token(RegistrationToken {
-                                    key: registroken_token,
-                                }),
+                                .borrow_mut(),
+                            TokenFactory::new(event.token.key),
                         )?;
                     }
                     PostAction::Remove => {
@@ -456,12 +463,12 @@ impl<'l, Data> EventLoop<'l, Data> {
                     let mut poll = self.handle.inner.poll.borrow_mut();
                     if let Err(e) = disp.unregister(
                         &mut poll,
-                        self.handle
+                        &mut self
+                            .handle
                             .inner
                             .sources_with_additional_lifecycle_events
-                            .create_register_for_token(RegistrationToken {
-                                key: registroken_token,
-                            }),
+                            .borrow_mut(),
+                        TokenFactory::new(registroken_token),
                     ) {
                         log::warn!(
                             "[calloop] Failed to unregister source from the polling system: {:?}",
diff --git a/src/sources/mod.rs b/src/sources/mod.rs
index 4ccc1d5..3111dfa 100644
--- a/src/sources/mod.rs
+++ b/src/sources/mod.rs
@@ -4,7 +4,7 @@ use std::{
     rc::Rc,
 };
 
-use crate::{sys::TokenFactory, Poll, Readiness, RegistrationToken, Token};
+use crate::{sys::TokenFactory, Poll, Readiness, Token};
 
 pub mod channel;
 #[cfg(feature = "executor")]
@@ -314,13 +314,13 @@ where
     fn register(
         &self,
         poll: &mut Poll,
-        mut additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
         token_factory: &mut TokenFactory,
     ) -> crate::Result<()> {
         let mut this = self.borrow_mut();
 
         if this.needs_additional_lifecycle_events {
-            additional_lifecycle_register.register();
+            additional_lifecycle_register.register(token_factory.key);
         }
         this.source.register(poll, token_factory)
     }
@@ -328,13 +328,13 @@ where
     fn reregister(
         &self,
         poll: &mut Poll,
-        mut additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
         token_factory: &mut TokenFactory,
     ) -> crate::Result<bool> {
         if let Ok(mut me) = self.try_borrow_mut() {
             me.source.reregister(poll, token_factory)?;
             if me.needs_additional_lifecycle_events {
-                additional_lifecycle_register.register();
+                additional_lifecycle_register.register(token_factory.key);
             }
             Ok(true)
         } else {
@@ -345,12 +345,13 @@ where
     fn unregister(
         &self,
         poll: &mut Poll,
-        mut additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
+        token_factory: TokenFactory,
     ) -> crate::Result<bool> {
         if let Ok(mut me) = self.try_borrow_mut() {
             me.source.unregister(poll)?;
             if me.needs_additional_lifecycle_events {
-                additional_lifecycle_register.unregister();
+                additional_lifecycle_register.unregister(token_factory.key);
             }
             Ok(true)
         } else {
@@ -382,60 +383,43 @@ pub(crate) trait EventDispatcher<Data> {
     fn register(
         &self,
         poll: &mut Poll,
-        additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
         token_factory: &mut TokenFactory,
     ) -> crate::Result<()>;
 
     fn reregister(
         &self,
         poll: &mut Poll,
-        additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
         token_factory: &mut TokenFactory,
     ) -> crate::Result<bool>;
 
     fn unregister(
         &self,
         poll: &mut Poll,
-        additional_lifecycle_register: AdditionalLifetimeEventsRegister<'_>,
+        additional_lifecycle_register: &mut AdditionalLifecycleEventsSet,
+        token_factory: TokenFactory,
     ) -> crate::Result<bool>;
 
     fn before_sleep(&self) -> crate::Result<Option<(Readiness, Token)>>;
     fn before_handle_events(&self, was_awoken: bool);
 }
 
+/// Track sources which require extra lifecycle events.
 #[derive(Default)]
-/// The list of events
-pub(crate) struct AdditionalLifetimeEventsSet {
+pub(crate) struct AdditionalLifecycleEventsSet {
     /// The list of events. The boolean indicates whether the source had an event
     /// - this is stored in this list because we need to know this value for each
     /// item at once - that is, to avoid allocating in the hot loop
-    pub(crate) values: RefCell<Vec<(RegistrationToken, bool)>>,
+    pub(crate) values: Vec<(usize, bool)>,
 }
-
-impl AdditionalLifetimeEventsSet {
-    pub(crate) fn create_register_for_token(
-        &self,
-        token: RegistrationToken,
-    ) -> AdditionalLifetimeEventsRegister {
-        AdditionalLifetimeEventsRegister { set: self, token }
-    }
-}
-
-/// To ensure that registering the events which need additional lifecycle events
-/// happens correctly, we do this within the implementations of `EventDispatcher`
-/// `AdditionalLifetimeEventsRegister` is a helper type to centralise this logic
-pub(crate) struct AdditionalLifetimeEventsRegister<'a> {
-    set: &'a AdditionalLifetimeEventsSet,
-    token: RegistrationToken,
-}
-
-impl<'a> AdditionalLifetimeEventsRegister<'a> {
-    fn register(&mut self) {
-        self.set.values.borrow_mut().push((self.token, false))
+impl AdditionalLifecycleEventsSet {
+    fn register(&mut self, token: usize) {
+        self.values.push((token, false))
     }
 
-    fn unregister(&mut self) {
-        self.set.values.borrow_mut().retain(|it| it.0 != self.token)
+    fn unregister(&mut self, token: usize) {
+        self.values.retain(|it| it.0 != token)
     }
 }
 

```

DJMcNab:
So there are three changes you're asking for there:
1) To make `AdditionalLifecycleEventsSet` not have the inner `RefCell`. The current design was intended to avoid doing the `borrow_mut` operations in the cases where they weren't needed, but that is only a small difference, as these shouldn't be on the hot path. This transformation hardly changes the code length, so I'll defer to @elinorbgr on this
1) To make the vector inside `AdditionalLifecycleEventsSet` not type-safe. There is definitely a slight verbosity cost, but I believe it's generally better to be more type safe than less type safe. Again, I'll defer to @elinorbgr
1) To use the key which is already present in TokenFactory, thus avoiding `AdditionalLifecycleEventsRegister`. I had avoided doing this because the key field on `TokenFactory` isn't public. I think there are two aspects to this:
   - Whether the key field on `TokenFactory` should be exposed outside of `sys.rs`. I personally that would be a layering violation, but I think it is also much cleaner for this PR. I think I will make this change (by adding a `pub(crate) fn registration_token(&self)->RegistrationToken` method on `TokenFactory`)
    - Whether this key field should be exposed in the public API. I probably would expose it, but I've not done much thinking on that, and I'm not planning to do so in this PR.

elinorbgr:
Just to make sure I didn't miss anything, point 1) is just about having the `RefCell` inside or outside of the `AdditionalLifecycleEventsSet` right? Given how small the difference is in terms of running, I think I'd prefer to have it outside, as it'd be more coherent with the rest of the contents of `LoopInner`.

Regarding 2/3, I think I like the way it is currently on the PR, keeping the type-safety and using the token from he available tokenfactory is good imo, especially given this concerns crate-internals anyway.

elinorbgr:
Also, `AdditionalLifetimeEventsSet` is still named using *lifetime* rather than *lifecycle*, I guess this one slipped through your attention? :sweat_smile:

kchibisov (approved):
I don't have concerns with this anymore. Thanks.

src/loop_logic.rs
```diff
@@ -315,8 +377,29 @@ impl<'l, Data> EventLoop<'l, Data> {
...
+                    }
+                }
+            }
+            for (source, has_event) in &*extra_lifecycle_sources.values {
+                if let Some(disp) = self.handle.inner.sources.borrow().get(source.key) {
+                    disp.before_handle_events(*has_event);
+                } else {
+                    unreachable!()
+                }
+            }
```
elinorbgr, line 399: Couldn't these two loops be fused into a single one?

And then, it looks like `has_event` is only used by those two loops, so it can probably be removed altogether from `extra_lifecycle_sources`, no?

src/sources/mod.rs
```diff
@@ -157,25 +164,28 @@ pub trait EventSource {
...
     fn unregister(&mut self, poll: &mut Poll) -> crate::Result<()>;
 
-    /// Notification that a polling session is going to start
+    /// Whether this source needs to be sent the [`EventSource::before_sleep`]
+    /// and [`EventSource::before_handle_events`] notifications. These are opt-in because
+    /// they require more expensive checks, and almost all sources will not need these notifications
+    const NEEDS_EXTRA_LIFECYCLE_EVENTS: bool = false;
+    /// Notification that a single `poll` is about to begin.
+    /// If this returns Ok(Some), polling will shortcircuit, and
+    /// your event handler will be called with the returned Token and Readiness
```
elinorbgr, line 173: The current implementation does still invoke poll with a timeout of 0 even if the `before_sleep()` method returned `Ok(Some(...))`, which means it is possible that poll generates another legit event. We need to document what happens in this case.

What the current implementation does is:
- `before_handle_events()` is invoked if a real event was generated, independently of what `before_sleep()` returned
- If a real event is generated in addition to the one from `before_sleep()`, then `dispatch_events()` is invoked twice. 

Imo this is the correct thing to do (to handle composite sources with multiple tokens), but this needs to be documented clearly.

src/sources/mod.rs
```diff
@@ -157,25 +164,28 @@ pub trait EventSource {
...
-    where
-        F: FnMut(Self::Event, &mut Self::Metadata) -> Self::Ret,
-    {
-        Ok(())
-    }
+    /// Use this to perform a cleanup before event handlers with arbitrary
+    /// code may run. This could be used to drop a lock obtained in
+    /// [`EventSource::before_sleep`]
+    #[allow(unused_variables)]
+    fn before_handle_events(&mut self, was_awoken: bool) {}
```
elinorbgr, line 188: I'm wondering if it might be relevant to pass the `(Token, Readiness)` couple as argument to this, instead of just a `was_awoken` bool. Again as a way to properly handle composite sources that may have multiple underlying FDs and tokens.

src/loop_logic.rs
DJMcNab: I don't believe so - we need to run all before_handle_events handlers, even those which didn't receive events

I suppose we could have two different handler methods for the different parts, so that we don't need to track was_awoken. That means that if a source needs was_awoken, it would have to track it itself. But that's fine

I'm fairly sure we do need two loops though

elinorbgr: But what about something like this?

```
for each extra_lifecycle_source {
    let has_event = /* loop over events to check if that source was woken up */;
    disp.before_handle_events(has_event);
}

File: src/sources/mod.rs (hunk 10 above)
Comment (elinorbgr) [reply src/sources/mod.rs:188]: To be more precise, I see two ways to achieve this:

- make the function take as argument `Option<(Token, Readiness)>`, and say that it is called once with `None` if no event is received, or once ore more times with `Some()` if events were triggered
- make the function take a `&dyn Iterator<Item=(Token, Readiness)>>` as argument, with that iterator being empty if no event was triggered

File: Cargo.toml [hunk 11]
```diff
@@ -24,7 +24,7 @@ bitflags = "1.2"
 futures-io = { version = "0.3.5", optional = true }
 io-lifetimes = "1.0.3"
 log = "0.4"
-nix = { version = "0.26", default-features = false, features = ["signal"], optional = true }
+nix = { version = "0.26", default-features = false, features = ["signal",], optional = true }
```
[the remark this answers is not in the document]
DJMcNab: I have an extension which formats TOML files. In 0364c1caafb9d3704728b190edf4206dadd6ed36, I reverted some edits made by this manually, but missed this comma

src/sources/mod.rs
```diff
@@ -168,24 +169,34 @@ pub trait EventSource {
...
+    /// Notification that a single `poll` is about to begin
     ///
     /// Use this to perform operations which must be done before polling,
     /// but which may conflict with other event handlers. For example,
-    /// if polling requires a lock to be taken.
+    /// if polling requires a lock to be taken
+    ///
+    /// If this returns Ok(Some), this will be treated as an event arriving in polling, and
+    /// your event handler will be called with the returned `Token` and `Readiness`.
+    /// Polling will however still occur, so additional events from this or other sources
```
elinorbgr, line 180: "Polling will however still occur, *but with a timeout of 0*

src/sources/mod.rs
```diff
@@ -168,24 +169,34 @@ pub trait EventSource {
...
         Ok(None)
     }
-    /// Notification that polling is completed, and the resulting events are
-    /// going to be executed.
+    /// Notification that polling is complete, and [`EventSource::process_events`] will
+    /// be called with the given events for this source. The iterator may be empty,
+    /// which indicates that no events were generated for this source
+    ///
+    /// Please note, the iterator will also include any synthetic event returned from
+    /// [`EventSource::before_sleep`]
```
elinorbgr, line 193: Do we actually want that?

The source can already internally track if it gave a synthetic event or not, and at least for the wayland socket, it'd be more useful in this method to know if other *real* events occurred during the polling.

In both cases the wayland source will be able to work with it, but it seems to me including the synthetic events in this iterator only increases the complexity of implementing the sources.

elinorbgr (approved):

elinorbgr:
Alright, thanks a lot this is perfect! :+1:

PR #149, opened 2023-09-16
Fix the additional lifecycle events to work as documented

DJMcNab:
Required for https://github.com/Smithay/calloop-wayland-source/pull/1.

I think this only needs to be a patch release, as this just brings the behaviour in-line with the docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 #### Bugfixes
-- Make the additional lifecycle events match the docs
+- Fix `EventSource::before_handle_events()` being erroneously give an iterator over synthetic events instead of real events
 
 ## 0.12.0 -- 2023-09-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+#### Bugfixes
+- Make the additional lifecycle events match the docs
+
 ## 0.12.0 -- 2023-09-11
 
 #### Breaking changes
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -387,7 +387,7 @@ impl<'l, Data> EventLoop<'l, Data> {
                 for source in &mut *extra_lifecycle_sources.values {
                     if let Some(disp) = self.handle.inner.sources.borrow().get(source.key) {
                         let iter = EventIterator {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+#### Bugfixes
+- Make the additional lifecycle events match the docs
```
kchibisov, line 6: It's a bit unclear what was exactly changed, probably should describe the behavior that could be observed (like endless loop due to use of lifecycle events).

kchibisov (changes requested):
it's an internal change, so could be released as a patch release normally.

kchibisov:
CI should be fixed after https://github.com/Smithay/calloop/pull/150

CHANGELOG.md
DJMcNab: I think this is sufficient because:
1) there are, to my knowledge, no users of this API, so no-one will have run into this
2) The pr is readily available
3) I don't want to make the changelog be really long with useless content

But I'll put whatever you prefer, if you give me the text you want to put here

kchibisov: I mean, `- Fix infinite loop when using registering a source with lifecycle events` is short enough from what I can say? Or is it not reflecting the situation correctly?

DJMcNab: So the infinite loop was only because the events were never getting read - we were still making progress, and other sources would have worked.

That is, the infinite loop is in the WaylandSource - yes because of this issue, but that isn't the only way this could manifest

kchibisov: Hm, yeah, that's actually fair. I'm fine with the changelog like that, it's just reads without providing a real value to the user.

kchibisov: probably  @elinorbgr has a better suggestion on that matter, since we'd need to patch release and maybe yank 0.12.

elinorbgr: Overall, I'd prefer to have a quick description of what the bug was, so something like this would be fine imo:

```
- Fix `EventSource::before_handle_events()` being erroneously give an iterator over synthetic events instead of real events
```

elinorbgr:
:+1: with a rebase & changelog adjusment

elinorbgr:
Perfect thanks!

PR #155, opened 2023-10-03
Fix crash when dropping removed token

chrisduerr:
This fixes an issue where calloop would crash internally when the `PostAction::Drop` was invoked after the user manually cancelled the source in the callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Bugfixes
+
+- Fixed a crash when removing a token in a callback with a `PostAction::Remove`
+
 ## 0.12.2 -- 2023-09-25
 
 #### Bugfixes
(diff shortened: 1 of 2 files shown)


elinorbgr:
Could you give a quick summary of what you changed and why? This diff is not very clear to read...

chrisduerr:
> Could you give a quick summary of what you changed and why? This diff is not very clear to read...

Just changed the remove to try_remove. Just also tried to make the code somewhat readable at least.

kchibisov:
This probably needs a test given that it sounds like simple to repro?

elinorbgr:
Alright, I took the time to figure out exactly what happens, and this seems like a good way to fix this, thanks.

Could you please just add a changelog entry for this under the "unreleased" section?

elinorbgr:
Ok so, it appears that a deeper issue with this PR as a special case was found & fixed in #156.

elinorbgr:
closing in favor of #156

PR #156, opened 2023-10-04
Delay freeing Slab ids until events are processed

Drakulix:
Seems to solve #155 for me as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Bugfixes
+
+- Fix an issue, where id-reuse could execute a PostAction on a newly registered event source
+
 ## 0.12.2 -- 2023-09-25
 
 #### Bugfixes
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -69,7 +69,9 @@ impl RegistrationToken {
 
 pub(crate) struct LoopInner<'l, Data> {
     pub(crate) poll: RefCell<Poll>,
-    pub(crate) sources: RefCell<Slab<Rc<dyn EventDispatcher<Data> + 'l>>>,
+    // The `Option` is used to keep slots of the slab occipied, to prevent id reuse
+    // while in-flight events might still referr to a recently destroyed event source.
+    pub(crate) sources: RefCell<Slab<Option<Rc<dyn EventDispatcher<Data> + 'l>>>>,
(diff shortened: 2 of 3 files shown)


src/loop_logic.rs
```diff
@@ -69,7 +69,7 @@ impl RegistrationToken {
 
 pub(crate) struct LoopInner<'l, Data> {
     pub(crate) poll: RefCell<Poll>,
-    pub(crate) sources: RefCell<Slab<Rc<dyn EventDispatcher<Data> + 'l>>>,
+    pub(crate) sources: RefCell<Slab<Option<Rc<dyn EventDispatcher<Data> + 'l>>>>,
```
elinorbgr, line 72: lets add a comment here explaining the meaning and role of this `Option`

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix an issue, where id-reuse could execute a PostAction on a newly registered event source
+
```
elinorbgr, line 6: Could you put it in a "Bugfixes" subsection? To fit the organization of the previous releases.

elinorbgr:
Could you also integrate the test case from #155 in the PR, to make sure we're not missing anything?

elinorbgr:
Also, I'm not sure whether the CI failures on macos & freebsd are related to this PR or not, so I'll need to confirm this.

elinorbgr:
Ok, the current master CI does pass on those platform, so those are not flukes, looks like this somehow interacted with some subtle platform-specific behavior.

src/loop_logic.rs
```diff
@@ -477,17 +479,20 @@ impl<'l, Data> EventLoop<'l, Data> {
                             .inner
                             .sources
                             .borrow_mut()
-                            .remove(registroken_token);
+                            .get_mut(registroken_token)
+                            .take();
```
chrisduerr, line 483: Why are you not using `try_remove` here?

src/loop_logic.rs
Drakulix: Is this sufficient?

src/loop_logic.rs
Drakulix: Because we want to intentionally leave `None` in place, while there are still events being processed. I added some more comments to explain this.

elinorbgr: Ironically, this is exactly where the issue causing the CI failure is:

```suggestion
                            .get_mut(registroken_token)
                            .unwrap_or(&mut None)
                            .take();
```

PR #165, opened 2023-12-21
fix borrow error when scheduling future inside an executor's callback

i509VCB:


notgull (approved):
LGTM as the author of this code

elinorbgr:
Thanks!

PR #168, opened 2024-01-13
Add Windows support

### LINKED ISSUE #161: Windows support
This is likely a giant undertaking, but this crate looks amazing. But I can't lock my ui library into not working on Windows. I use Linux daily but you still have to use libraries that work on every platform.

notgull:
This commit adds Windows support to this package. Since polling already
has Windows support through IOCP, the main obstacle was adding a ping
event source using IOCP. The hardest part is emulating a pipe using some
shared state and a posted completion packet.

Fixes #161

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -37,10 +37,18 @@ use crate::{AdditionalLifecycleEventsSet, RegistrationToken};
 /// `AsyncWrite` if the underlying type implements `Read` and/or `Write`.
 ///
 /// Note that this adapter and the futures procuded from it and *not* threadsafe.
+/// 
+/// ## Platform-Specific
+/// 
+/// - **Windows:** Usually, on drop, the file descriptor is set back to its previous status.
+///   For example, if the file was previously nonblocking it will be set to nonblocking, and
+///   if the file was blocking it will be set to blocking. However, on Windows, it is impossible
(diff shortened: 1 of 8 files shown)


src/io.rs
```diff
@@ -165,9 +175,9 @@ impl<'l, F: AsFd> Drop for Async<'l, F> {
     fn drop(&mut self) {
         self.inner.kill(&self.dispatcher);
         // restore flags
-        let _ = fcntl_setfl(
+        let _ = set_nonblocking(
             unsafe { BorrowedFd::borrow_raw(self.dispatcher.borrow().fd) },
-            self.old_flags,
+            false,
```
elinorbgr, line 180: This is a slight change in behavior. Currently, if the the FD was already in non-block mode before being wrapped into an `Async<_>`, this is preserved on drop. While with this change, the FD is unconditionally set to blocking mode.

notgull: The thing with Windows is that it's impossible to tell whether or not a socket is currently in nonblocking mode. The socket can be set to blocking or non-blocking mode, but there is no way to tell what it currently is (without trying to perform a read of course).

So this behavior can't really be ported to Windows. I think the best course of action is to keep the current behavior on Unix, but on Windows to only ever set it to blocking mode.

elinorbgr (approved):

Issue #171, opened 2024-01-29
Not all tasks scheduled with `Executor` are run


ids1024:
Not sure exactly what's going on, but some code I have trying to use calloops `Executor` doesn't always end up running the scheduled tasks. Adding a print to the start of the `async {` block shows it never starts. Using a different executor doesn't have this problem.

Here's a minimal case where I see an issue like this. It's not really reflective of exactly what I was doing where I ran into this, but it's an interesting case, where `"Baz"` surely should be printed, but isn't:

```rust
fn main() {
    let (executor, scheduler) = calloop::futures::executor().unwrap();
    let scheduler_clone = scheduler.clone();
    scheduler.schedule(async move {
        println!("Foo");
        scheduler_clone.schedule(async {
            println!("Bar");
        });
    });
    let mut event_loop =  calloop::EventLoop::try_new().unwrap();
    event_loop.handle().insert_source(executor, |_: (), _, _| {}).unwrap();
    event_loop.run(None::<std::time::Duration>, &mut (), |_| {
        scheduler.schedule(async {
            println!("Baz");
        });
    });
}
```

ids1024:
It looks like in both cases, removing `if self.notified.swap(true, Ordering::SeqCst) { return; }` fixes this. Though that logic seems reasonable. Hm.

ids1024:
Ah, so if `schedule` is called in the callback invoked by `process_events`, that will set `notified` to `true` (which had been cleared to `false` at the start of `process_events`, and send a ping. But the `if clear_readiness { ... }` block results in that ping being ignored, but doesn't set `notified` to `false`.

PR #172, opened 2024-01-29
executor: Fix issue when task is scheduled within event callback

ids1024:
I believe this should be overall more correct. Previously, if the callback taken by `process_events` scheduled a task, `process_events` could exit with the readiness of `self.ping` cleared, but `notified` set to `true`. In which case, future calls to `schedule` would not wake the event source.

Fixes https://github.com/Smithay/calloop/issues/171.


notgull (approved):
Thanks!

PR #182, opened 2024-04-25
Remove nix from the public API

### LINKED ISSUE #179: Make `nix` not a public dependency
I see 0.13.0 is a fairly minor change, but is a semver bump because `nix` is a public dependency. Just for the `Signal` and `siginfo` type rexports used with the `signals` feature it seems? That doesn't seem ideal.

It may also be better not to use `nix`, but that is less important than removing it from the public API.

Maybe `Signal` could just be a `c_int`, then the caller could easily use the constants in `libc`? Not sure about `siginfo`. A struct could be defined in calloop providing certain fields.

notgull:
This commit removes nix from the public API. nix has frequent breaking
changes due to the large API surface it covers, which makes it a semver
hazard to keep it in the public API.

Thus far only two items from nix are exposed: Signal and siginfo. Signal
is an enum consisting of most real-life Unix signals. I have replaced it
with a hand rolled enum primarily taken from the one in async-signal.
For the other case, the Event structure exposes most of the relevant
fields. So I've turned these fields into accessors. We can PR more
later.

This is a breaking change, but it's in service of avoiding breaking
changes in the future.

Closes #179


elinorbgr (approved):
Yeah, lgtm. Nix in public API has long be a pain.

ids1024 (approved):

PR #184, opened 2024-05-17
Optimize for timer timeout extension

cmeissl:


notgull:
@cmeissl The CI issues are fixed on master, please rebase

notgull:
@cmeissl try again

cmeissl:
> @cmeissl try again

The CI failed because the version (latest) of `criterion` depends on `clap` which requires a newer MSRV. I downgraded `criterion` for now, but maybe it would also be an option to split the tests and the MSRV checks in the future.

cmeissl:
@notgull I had to downgrade a few other deps, but the CI is now green finally!

notgull (approved):
Thanks!

PR #185, opened 2024-05-19
Fix nightly clippy warning

notgull:


notgull:
Confirmed that the failing test is `test loop_logic::tests::insert_bad_source`, which appears to only fail on nightly.

Issue #194, opened 2024-06-21
tracing vs log


joshka:
I noticed when reading the code, there's a lot of `log::trace!("[calloop] some message)` calls.
I wonder if you'd be ok with moving the lib to just use tracing (with a few small changes for structured values being reported).

Some things of note:
- tracing is a dependency of the polling crate, so it will be in the dependency tree already
- tracing is a dependency of the biggest dependant of calloop by downloads (winit)
- Tracing brings in proc-macro2 and syn. I saw you removed thiserror recently with the aim to remove these crates, but these are deps of the polling crate also (due to tracing).
- tracing can write to log if needed.

PR incoming that aligns with this.

PR #195, opened 2024-06-21
chore: use tracing instead of log crate

joshka:
Fixes: https://github.com/Smithay/calloop/issues/194

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,11 +22,11 @@ codecov = { repository = "Smithay/calloop" }
 async-task = { version = "4.4.0", optional = true }
 bitflags = "2.4"
 futures-io = { version = "0.3.5", optional = true }
-log = "0.4"
 pin-utils = { version = "0.1.0", optional = true }
 polling = "3.0.0"
-slab = "0.4.8"
 rustix = { version = "0.38", default-features = false, features = ["event", "fs", "pipe", "std"] }
+slab = "0.4.8"
+tracing = { version = "0.1.40", default-features = false, features = ["log"] }
 
 [target.'cfg(unix)'.dependencies]
 nix = { version = "0.29", default-features = false, features = ["signal"], optional = true }
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -14,8 +14,8 @@ use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 #[cfg(windows)]
 use std::os::windows::io::{AsHandle, AsRawHandle, AsSocket as AsFd, BorrowedHandle, RawHandle};
 
-use log::trace;
 use polling::Poller;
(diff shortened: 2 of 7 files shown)


joshka:
Tracing logs for a sample app that hits these - 
image

notgull (commented):
My only hesitancy is for the long-term maintenance of the tracing crate. I know that there was a period where tracing was effectively unmaintained. Looking at the repo it looks like it's maintained better now, but I'm just worried about needing to rollback to `log` if trading ever becomes unmaintained again.

Looks good to me otherwise.

joshka:
Fixed a log::warn that I'd missed. The CI still breaks due to the log downgrade. The MSRV of the current version of log (1.60.0) seems to be lower than calloop's MSRV (1.63.0). Does this mean that this is not necessary any more?

> My only hesitancy is for the long-term maintenance of the tracing crate. I know that there was a period where tracing was effectively unmaintained. Looking at the repo it looks like it's maintained better now, but I'm just worried about needing to rollback to `log` if trading ever becomes unmaintained again.

The other thing I noted (in #194 in case you missed it) is that this brings back in the proc-macro2 and syn crates to the dependency list (which I saw you were trying to remove by avoiding `thiserror`).

notgull:
> The CI still breaks due to the log downgrade. The MSRV of the current version of log (1.60.0) seems to be lower than calloop's MSRV (1.63.0). Does this mean that this is not necessary any more?

Yes, it's probably not necessary.

> The other thing I noted (in #194 in case you missed it) is that this brings back in the proc-macro2 and syn crates to the dependency list (which I saw you were trying to remove by avoiding `thiserror`).

If you import tracing without default features it doesn't have these dependencies though, right?

joshka:
> If you import tracing without default features it doesn't have these dependencies though, right?

Actually yep, you're right - I must have checked before I added that

```
calloop v0.14.0 (/Users/joshka/local/calloop)
├── bitflags v2.5.0
├── polling v3.7.2
│   ├── cfg-if v1.0.0
│   ├── rustix v0.38.34
│   │   ├── bitflags v2.5.0
│   │   ├── errno v0.3.9
│   │   │   └── libc v0.2.155
│   │   └── libc v0.2.155
│   └── tracing v0.1.40
│       ├── log v0.4.21
│       ├── pin-project-lite v0.2.14
│       └── tracing-core v0.1.32
├── rustix v0.38.34 (*)
├── slab v0.4.9
│   [build-dependencies]
│   └── autocfg v1.3.0
└── tracing v0.1.40 (*)
```

notgull:
Please rebase on `master` to fix the CI issues. Then we should be good to go.

joshka:
> Please rebase on `master` to fix the CI issues. Then we should be good to go.

Done

notgull (approved):
Thanks!

PR #200, opened 2024-07-21
Bound channel elements processed every time process_events is called

notgull:
At the moment, the channel implentation processes as many channel
elements as possible every time "process_events" is called. However,
in multithreaded cases this can cause the loop to be stuck in the
"process_events" section of a channel forever. If one thread keeps
sending new elements into the channel while the current thread keeps
reading them, it will starve other event sources of running time.

This commit fixes this issue by bounding the number of channel elements
that can be processed every time "process_events" is called. It chooses
the smallest of the following numbers:

- The capacity of the channel.
- 1024 (chosen because this is also used by async-executor)

If the channel is not empty after we have read this number of elements,
the underlying source is not triggered. This should make it so the
channel is immediately re-polled on the next dispatch. However it gives
other sources more time to run.


notgull:
It appears this has exposed a bug in Windows' `Ping` implementation. Standby while I look into this.

PR #201, opened 2024-07-21
Remove unmaintained actions-rs actions

notgull:
Also, uses taiki-e/checkout-actions instead of actions/checkout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout sources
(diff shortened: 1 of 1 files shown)


YaLTeR:
Fwiw there's now https://github.com/dtolnay/rust-toolchain which deals with some annoying edge cases

notgull:
> Fwiw there's now https://github.com/dtolnay/rust-toolchain which deals with some annoying edge cases

I don't think we run into any of these edge cases on GitHub hosted runners, with our configuration. So I think we're fine.

PR #204, opened 2024-07-29
Make `Generic` `UnwindSafe` even if its error type isn't

dfoxfranke:
`std::io::Error` does not implement `UnwindSafe` on account of its ability to wrap a `Box<dyn Error + Send + Sync>`. This infects `Generic` which in turn infects `Channel`, which is inconvenient because it means you can't construct a channel pair, pass the receiver side to a child thread, and then handles panics in the child.

But this is silly, because the error parameter to `Generic` is just a phantom. We aren't really wrapping an error at all, let alone downcasting it to a mutable reference and mucking with the referent. This patch changes the `PhantomData<E>` to `PhantomData<AssertUnwindSafe<E>>` so that `Generic<F, E>` can be `UnwindSafe` as long as `F` is.

### DIFF
--- a/src/sources/generic.rs
+++ b/src/sources/generic.rs
@@ -42,7 +42,7 @@
 //! [`EventSource`](crate::EventSource) implementation to them.
 
 use polling::Poller;
-use std::{borrow, marker::PhantomData, ops, sync::Arc};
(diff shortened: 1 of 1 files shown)


notgull (approved):
Thanks!

PR #205, opened 2024-07-31
On same deadline, items queued first have priority

amcandio:
When multiple are registered at the same time, those registered first should have priority to make scheduling behavior easier to predict. I don't think this should be added as documented behavior to prevent users from depending on it.

One nice to have follow up would be to start supporting custom prioritization as tie-breaker so users can define their own to handle tie-breaking. Right now users depend on tweaking expire timestamps to get same effect, which is not clean.

### DIFF
--- a/src/sources/timer.rs
+++ b/src/sources/timer.rs
@@ -278,7 +278,8 @@ impl TimerWheel {
 impl std::cmp::Ord for TimeoutData {
     #[inline]
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        // earlier values have priority, on same deadline items queued first have priority
+        // Earlier values have priority.
+        // For items with the same deadline, those queued first have priority.
         self.deadline
             .cmp(&other.deadline)
             .then_with(|| self.counter.cmp(&other.counter))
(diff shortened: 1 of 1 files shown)


notgull:
@amcandio Please rebase on `master` to fix the CI

amcandio:
@notgull rebased!

amcandio:
@notgull any update on this?

src/sources/timer.rs
```diff
@@ -278,8 +278,11 @@ impl TimerWheel {
 impl std::cmp::Ord for TimeoutData {
     #[inline]
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        // earlier values have priority
-        self.deadline.cmp(&other.deadline).reverse()
+        // earlier values have priority, on same deadline items queued first have priority
```
notgull, line 281: Please use proper capitalization and punctuation for comments.

notgull (changes requested):
Apologies, I was distracted.

amcandio:
@notgull apologies for the delay!

amcandio:
Hi! Any chance we can ship this one? Not super important but I have this old tech debt item that I could close with this feature!

PR #208, opened 2024-09-06
channel: Return `PostAction::Remove` on disconnect

ids1024:
It seems 86c6713 broke this. By maintaining a copy of `Ping`, the ping source will never return `PostAction::Remove`.

(I guess we could use more tests to catch things like this...)


notgull (approved):
Thanks!

Issue #203, opened 2024-07-25
More Readiness events?


vikigenius:
I see that Mio has support for:

[is_read_closed](https://docs.rs/mio/latest/mio/event/struct.Event.html#method.is_read_closed)

[is_write_closed](https://docs.rs/mio/latest/mio/event/struct.Event.html#method.is_write_closed)

Can we get that in Calloop as well?

I am trying to use a client server architecture with Unix Sockets. Multiple clients can connect to the server. In rust every time there is a new connection to UnixListener it get's it's own UnixStream. I would like to keep track of connections so that I can send messages to appropriate streams.

But I also need to detect when the stream has been shutdown and we know that we won't receive anymore data in it.

Is there a reason why these aren't implemented? It seems like Mio does it using EPOLLHUP, EPOLLIN, EPOLLRDHUP.
Is there a way to achieve this in calloop already that I don't know about?

notgull:
`polling` has support for `is_interrupt` (usually corresponds to HUP), which I think fits the bill here.

vikigenius:
@notgull Seems like it, looks like it maintains an extra field

```rust
/// Indicates that a file descriptor or socket can read or write without blocking.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct Event {
    /// Key identifying the file descriptor or socket.
    pub key: usize,
    /// Can it do a read operation without blocking?
    pub readable: bool,
    /// Can it do a write operation without blocking?
    pub writable: bool,
    /// System-specific event data.
    extra: sys::EventExtra,
}
```

Can calloop do something similar by maintaining an extra field in the Readiness struct and implementing is_interrupt using that? I am willing to contribute if this is something that you are interested in? I don't think this should be a breaking change?

notgull:
> I am willing to contribute if this is something that you are interested in?

Sorry for the delay. Yes I would be interested.

PR #210, opened 2024-09-16
Added readiness and interest for interrupt

vikigenius:
Fixes https://github.com/Smithay/calloop/issues/203


notgull (commented):
Note to my future self: this is a breaking change.

notgull:
This overall looks good to me, by the way. I'm just considering if I want to bundle this with any other breaking changes. I was planning for the next release of `calloop` to be `calloop` v1.0.

PR #213, opened 2024-11-20
Remove `mut` from `Poll`

Paraworker:
The `Poll` already has interior mutability, so wrap it with `RefCell` is unnecessary. This is a breaking change though.


notgull:
I wouldn't take this, since I believe it is a breaking change for no good reason.

PR #214, opened 2024-11-20
refactor: Simplify timeout calculation in Poll::poll()

kenoss:
Refactor using `Instant::saturating_duration_since()` and `Option::or()`.

### DIFF
--- a/src/sys.rs
+++ b/src/sys.rs
@@ -222,14 +222,14 @@ impl Poll {
 
     pub(crate) fn poll(&self, mut timeout: Option<Duration>) -> crate::Result<Vec<PollEvent>> {
         // Adjust the timeout for the timers.
-        let timeout2 = self
+        let next_timeout = self
             .timers
             .borrow()
             .next_deadline()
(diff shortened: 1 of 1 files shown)


src/sys.rs
```diff
@@ -215,17 +221,15 @@ impl Poll {
...
         // Adjust the timeout for the timers.
-        if let Some(next_timeout) = self.timers.borrow().next_deadline() {
-            if next_timeout <= now {
-                timeout = Some(Duration::ZERO);
-            } else if let Some(deadline) = timeout {
-                timeout = Some(std::cmp::min(deadline, next_timeout - now));
-            } else {
-                timeout = Some(next_timeout - now);
-            }
+        let timeout2 = self
```
notgull, line 225: Change this name to "next_timeout"

notgull (changes requested):

src/sys.rs
kenoss: Thanks! Done.

notgull (approved):

PR #216, opened 2025-01-08
Add `WeakLoopHandle`

Paraworker:
Introduce `WeakLoopHandle` to avoid reference cycles.


src/loop_logic.rs
```diff
@@ -288,6 +294,27 @@ impl<'l, Data> LoopHandle<'l, Data> {
     pub fn adapt_io<F: AsFd>(&self, fd: F) -> crate::Result<crate::io::Async<'l, F>> {
         crate::io::Async::new(self.inner.clone(), fd)
     }
+
+    /// Create a weak reference to this loop data.
```
notgull, line 298: Can you add an example for this function here?

notgull (changes requested):
One nit

src/loop_logic.rs
Paraworker: example added.

notgull (approved):
LGTM, rebase on master to fix the errors.

Paraworker:
Rebased.

notgull:
It looks like 1.63 CI is failing for some reason

Paraworker:
> It looks like 1.63 CI is failing for some reason

Looks like it requires `T: Debug` for `Weak<T>` in 1.63, hmm...
Implemented `Debug` for `WeakLoopHandle` manually now.

notgull (approved):
Thanks!

PR #219, opened 2025-01-12
Implement `Clone` and `Default` for `WeakLoopHandle` manually

Paraworker:
I suddenly realized this...

The derived `Clone`/`Default` implementation adds `Clone`/`Default` bound on the generic parameter `Data`.

### DIFF
--- a/src/loop_logic.rs
+++ b/src/loop_logic.rs
@@ -67,7 +67,6 @@ pub struct LoopHandle<'l, Data> {
 }
 
 /// Weak variant of a [`LoopHandle`]
-#[derive(Clone, Default)]
 pub struct WeakLoopHandle<'l, Data> {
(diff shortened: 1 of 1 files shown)


src/loop_logic.rs
```diff
@@ -332,6 +332,15 @@ impl<Data> Debug for WeakLoopHandle<'_, Data> {
     }
 }
 
+impl<Data> Clone for WeakLoopHandle<'_, Data> {
```
ids1024, line 335: Might be good to add a comment mentioning why it isn't just derived.

notgull: Agreed

Paraworker: Comments added, also changed `Default` to manual.

notgull (approved):

PR #227, opened 2025-06-17
source/futures: Fix handling of `notified` atomic

ids1024:
Previously, `notified` was cleared before the
`self.source.process_events` call. This meant that it was cleared before reading from the ping source eventfd/pipe.

This means that a call to `send()` form a waker in a different thread could call `ping()` before this has been read, and prevent future calls to `ping()`. But `process_events()` won't be woken again, freezing the executor.

This seems to fix an issue I've seen in `cosmic-workspaces` when trying to use the `calloop` executor.

### DIFF
--- a/src/sources/futures.rs
+++ b/src/sources/futures.rs
@@ -309,15 +309,15 @@ impl<T> EventSource for Executor<T> {
     {
         let state = &self.state;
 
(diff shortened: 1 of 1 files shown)


notgull (approved):
Thanks!

